PR #2190, opened 2017-12-11
E2e against ssr

mxstbr:
This should help us catch more errors in our server-side rendering setup. (like infinite loops) This requires an update to our Codeship config, don't be surprised if the tests fail here at first!

This is also based on #2183, so until that's merged you'll see those changes in this branch too.

Closes #2132 

**After merging**

- [ ] Remove unnecessary steps from Codeship build pipeline


mxstbr:
Way out of date and replaced with #2843

PR #2632, opened 2018-03-21
Slack Bot

mxstbr:
- [x] Figure out permission scopes necessary
- [x] Figure out how to send messages (#2015)
- [x] Send messages on new threads from athena
- [x] Get members on click, rather than automatically
- [x] Build UI in community settings
  - [x] Refactor pane to handle more general purpose
  - [x] Separate import step
  - [x] Re-connecting with new permissions
  - [x] Build UI for webhooks
- [x] Figure out how to handle when we have a token with the old permissions
- [x] Check whether we need admin perms to load user emails
  - [ ] We don't, so how do we make sure a normal member can't spam all members of a team without prompting for admin access?

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- iris (api)
- hyperion (frontend)

- Slack bot! You can now get notified in your Slack team whenever a new thread is posted in your community on Spectrum


brianlovin:
Made a lot of progress here, but it's a bit slow going as I'm having to make quite a few changes to models as I go. I've successfully refactored away from storing `slackImports` as a record type and moved slack settings for a community to the `communitySettings` model - this will make managing these things much easier down the road.

Tomorrow I'll implement all the channel matching to pipe notifications from spectrum channels to slack channels; this will mostly be working with `channelSettings` records and introducing some new queries and mutations; should be fun.

@mxstbr one major change I decided to make here to dramatically simplify the UX is to remove the "import members" step from the client. It was very confusing from a messaging and user flow point of view to have the ability to "import members" and then "re-import members" if you waited a few days or something. Instead, now a person just authenticates their community to a slack team, and at any point they can click one button to "invite members" which kicks off an athena job to send emails + notifications. One side benefit of this is that we never store any members data in our db as the athena queue only makes one query to get that the moment the user clicks "invite"

One side effect of all of this is I'll need to write a migration (which needed to happen anyway) to get all the slack imports onto the `communitySettings` records, and then to make sure we don't store all the `members` data from previous slack imports.

mxstbr:
That's nice, I like it! :100:

brianlovin:
cc @mxstbr - I'm running into cases where I'm rewriting so much of the exact same code in every mutation resolver, and I'd like to abstract this. I don't want to blow up this entire PR and refactor every resolver, but I can start with just channel resolvers. Would love your feedback before I go too far:

1. https://github.com/withspectrum/spectrum/pull/2632/commits/2a97b3c399dd5dac42cd4f34e9d83cbc1d2ab8eb - implements a directive resolver that checks for an authed user in context. Removes the same 3 repeated first lines of basically *every* mutation resolver we have
2. Add consistent permission check utils - https://github.com/withspectrum/spectrum/pull/2632/commits/6759783d65192a2920a154433f9e9179cc42b2c8

These two combined changes turn a resolver like this (where a majority of the code here is the *exact* same logic required in multiple resolvers): 

```js
export default async (
  _: any,
  { input: { id: channelId } }: DisableChannelTokenJoinInput,
  { user, loaders }: GraphQLContext
) => {
  const currentUser = user;
  if (!currentUser) {
    return new UserError('You must be signed in to manage this channel.');
  }

  const [channelPermissions, channel, settings] = await Promise.all([
    loaders.userPermissionsInChannel.load([currentUser.id, channelId]),
    loaders.channel.load(channelId),
    loaders.channelSettings.load(channelId),
  ]);

  const communityPermissions = await loaders.userPermissionsInCommunity.load([
    currentUser.id,
    channel.communityId,
  ]);

  if (!channelPermissions || !communityPermissions) {
    return new UserError("You don't have permission to do this.");
  }

  const canEdit =
    channelPermissions.isOwner ||
    channelPermissions.isModerator ||
    communityPermissions.isOwner ||
    communityPermissions.isModerator;

  if (!canEdit) {
    return new UserError("You don't have permission to do this.");
  }

  loaders.channelSettings.clear(channelId);

  // settings.id tells us that a channelSettings record exists in the db
  if (settings.id) {
    return await disableChannelTokenJoin(channelId);
  } else {
    return await createChannelSettings(channelId);
  }
};
```

to:

```js
export default async (
  _: any,
  { input: { id: channelId } }: DisableChannelTokenJoinInput,
  { user, loaders }: GraphQLContext
) => {
  if (await !userCanManageChannel(user.id, channelId)) {
    return new UserError('You don’t have permission to manage this channel');
  }

  return await getOrCreateChannelSettings(channelId).then(
    async () => await disableChannelTokenJoin(channelId)
  );
};
```

Open to ideas here; I know there are some folks that argue against direct resolvers, saying the polute the schema, but I don't mind keeping all our validation rules in one place there :/

mxstbr:
@brianlovin love it! I think @andrewingram would hate that haha

api/models/channelSettings.js
```diff
@@ -104,3 +107,58 @@ export const resetChannelJoinToken = (id: string) => {
     .run()
     .then(async () => await getChannelById(id));
 };
+
+type UpdateInput = {
+  channelId: string,
+  slackChannelId: ?string,
+  eventType: 'THREAD_CREATED',
```
mxstbr: Why not just make this `eventType: 'threadCreated'`? Then you don't have to do that awkward switch further below to match `threadCreated` to `THREAD_CREATED`, that seems pretty unnecessary.

api/mutations/channel/archiveChannel.js
```diff
@@ -1,36 +1,19 @@
...
+  { user, loaders }: GraphQLContext
 ) => {
-  const currentUser = user;
-
-  // user must be authed to delete a channel
-  if (!currentUser) {
-    return new UserError(
-      'You must be signed in to make changes to this channel.'
-    );
+  if (await !userCanManageChannel(user.id, channelId)) {
```
mxstbr: It'd be better if these lived in the GraphQL context so you

a) wouldn' have to pass the user ID manually every time
b) **you could use the loaders (properly)**

```JS
const resolver = (root, args, { permissions }) => {
  if (!(await permissions.canManageChannel(channelId)) { /* ... */ }
}
```

This would be implemented something like this:

```JS
const createPermissionsMethods = (userId: string, loaders) => ({
  canManageChannel: (channelId: string) => { return loaders.userChannelPermissions.load(userId, channelId) }
})

graphqlExpress(req => {
  const loaders = createLoaders();
  return {
    context: {
     loaders,
     permissions: createPermissionsMethods(req.user && req.user.id, loaders),
    }
  }
})
```

Alternatively we could also overload these onto the `user` key, which might be even clearer than `permissions`:

```JS
const resolver = (root, args, { user }) => {
  if (!(await user.canManageChannel(channelId)) { /* ... */ }
}
```

api/mutations/channel/utils.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+import createLoaders from '../../loaders/';
+const loaders = createLoaders();
```
mxstbr: **IMPORTANT: Never ever do this!!!**

Loaders are meant to be per-request, and they cache loaded data for their lifetime. If you do this, you'll get stale data after the first load!!!!

api/mutations/channel/utils.js
```diff
@@ -0,0 +1,38 @@
...
+
+  if (!communityPermissions) return false;
+  if (communityPermissions.isOwner || communityPermissions.isModerator)
+    return true;
+  if (!channelPermissions) return false;
+  if (channelPermissions.isOwner || channelPermissions.isModerator) return true;
+  return false;
+};
+
+export const userCanCreateChannel = async (
```
mxstbr: Rather than tying this to a specific mutation why don't we call this `userOwnsCommunity`? (which is really what it checks) That way we can reuse it for all the community mutations too.

I don't think it makes sense to externalise all permissions handling outside of the resolvers, I think instead these should be used as helper methods inside of the resolvers themselves.

mxstbr: To elaborate a bit here since this might not be immediately obvious:

```JS
// mutations/channel/createChannel.js
const resolver = (/* ... */) => {
  if (!(await userCanCreateChannel(input.communityId))) { /* ... */ }
}

// vs.

const resolver = (/* ... */) => {
  if (!(await userOwnsCommunity(input.communityId))) { /* ... */ }
}
```

The first one externalises permissions handling and requires you to go look up a different file to figure out when you're allowed or not to create a channel. This indirection is really unnecessary, as the `userCanCreateChannel` method won't ever be used anywhere.

The second one looks similar on the surface, but is very different in that it only adds some nice helper sugar on top of our permissions system. Importantly, it doesn't externalise the permissions handling at all, it's all still in the resolver and doesn't require me to look up a different file to figure out when I'm allowed to create a channel.

AndrewIngram: @mxstbr FWIW, I just think directives are the wrong solution to the problem of "I have too much repetitious logic".

Using Graphene at the moment, so I don't have a Node example, but this is how we're handling fields that require auth:

```python
@anonymous_return(list)
def resolve_foo(root, info, **args):
    return whatevs
```

The decorator accepts a direct value or a callable, and returns that if you attempt to resolve the field without an authenticated user.

This has nice advantages, because it's super-easy return things like empty arrays, nulls, zero values or whatever, rather than just a yes/no check of whether anonymous users are allowed to do something. I also think it's wrong for a GraphQL server to *ever* throw when executing a query unless something is actually broken. All user-facing error outcomes should be modelled into the schema.

I use this on both regular fields and mutations. In the case of mutations, I always use unions to represent success/failure payloads, so I tell anonymous_return to immediately return the failure payload:

```python
@anonymous_return(CreateFooFailure(message='You must be logged in to create a foo'))
def resolve_create_foo(root, info, **args):
    return whatevs
```

mxstbr: @AndrewIngram Appreciate the note! This specific comment is not about directives, these are just helper utils we use in the resolvers themselves. 

```JS
const resolver = (/* ... */) => {
  if (!(await userOwnsCommunity(input.communityId))) { /* ... */ }
}
```

The only directive that was added here is `isAuthed`, which only checks to make sure that a user is authenticated before running mutations—we won't move the permissions stuff to directives I'm pretty sure. :blush: 

> I also think it's wrong for a GraphQL server to ever throw when executing a query unless something is actually broken. All user-facing error outcomes should be modelled into the schema.

This is interesting, because we definitely don't do that right now. We have a `UserError` class which we throw when we want the error message to be exposed to the user, and any other error we send a generic `Internal server error` and log it to Sentry. What's the benefit of modelling that in the schema itself?

AndrewIngram: If you want to show a message to a user, you already have a mechanism for defining a contract between client and server -- it's your GraphQL schema. By moving user-facing data outside of the schema and into the errors array, you're back into the world of ad-hoc data structures and implicit conventions.

When you have a powerful expressive type system (GraphQL's is no Elm or Haskell, but it's decent), errors-as-data is beautifully elegant way of working, and pairs really nicely with functional programming. Errors that interrupt program flow (throwing exceptions that only get caught by a top-level) aren't as nice to work with, so should really be for the "something has unexpectedly gone wrong" scenario -- the equivalent of a 500 error.

You've seen all the Elm talks about making impossible states impossible, and allowing your types to force you to handle empty/loading/error states. This is the power of a comprehensive type system, it creates more work, but results in better software.

mxstbr: Do you have a specific example of what that would look like in the schema? Do you just have a generic Error type?

```GraphQL
type Thread {
  author: User | Error
}
```

and then from the frontend you do something like

```JS
if (props.data.error) {
  // render error state
}
```

if I understand that correctly?

AndrewIngram: I have a mixture of generic and non-generic errors, let's look at a deliberately complex example.

Say I have a mutation for a form, let's call it `addUser`, and you need to be both authenticated and have certain permissions, the client would handle anonymous users differently from users without permission (show login form versus show an access denied message). I might choose to model it this way:

```graphql
type LoginRequired {
  message: String!
}

type InsufficientPermissions {
  message: String!
}

type FieldError {
  fieldName: String!
  messages: [String!]! 
}

type FormValidationFailure {
   fieldErrors: [FieldErrors!]!
   nonFieldErrors: [String!]!
}

type AddUserSuccess {
   user: User!
   ... other fields
}

union AddUserPayload = LoginRequired
 | InsufficientPermissions 
 | FormValidationFailure 
 | AddUserSuccess

type mutation {
  ...
  addUser(
    username: String!, 
    email: String!, 
    password: String!
  ): AddUserPayload!
  ...
}
```

It might be a bit overwhelming, but hopefully you can see that the `AddUserPayload` union is capable of modelling a range of scenarios from super-generic, to super-specific. Some of the types will be shared with other mutations and query field, some might be specific to the mutation at hand. Similar to how your React components will vary widely in the degree to which they get reused, your error modelling can work in the same way.

api/mutations/channel/utils.js
brianlovin: Actually did this intentionally to keep permissions separate from results, based on us wanting to move to use loaders in mutations. This is already a problem in the branded login mutation resolvers where we have to remember to do `loaders.foo.clear(id)` . Honestly I'd prefer to just never use loaders in mutations, just because it's the highest likelihood of returning stale data.

So, I don't mind removing this, but I'm not opposed to the idea of permissions checks having their own caching via loaders. We can discuss further monday.

brianlovin:
Added a migration that for now just removes all member data from our db, which is a good thing. But in the future we can drop the slack imports table entirely, let's just hang on to it for a while just in case we need to check anything after shipping this.

brianlovin:
Regarding the permissions conversation above:

> Rather than tying this to a specific mutation why don't we call this userOwnsCommunity? (which is really what it checks) That way we can reuse it for all the community mutations too.

Because we allow moderators to create channels as well (fixed in https://github.com/withspectrum/spectrum/pull/2632/commits/2bd603de991d76450544ee32f452ce996d314805) - there's no reason to do `userIs{Role}InCommunity` because it's way more complex than doing `{ isOwner } = loaders.usersPermissionsInCommunity.load([userId, communityId])` (for example). 

The reason I have `userCanCreateChannel` is because the abstraction is more specific (can do this particular thing) and encompasses two permissions (owner and moderator) who can do the thing. In the future this may also be expanded...

> Alternatively we could also overload these onto the user key, which might be even clearer than permissions:

That pattern looks pretty nice actually :)

> It might be a bit overwhelming, but hopefully you can see that the AddUserPayload union is capable of modelling a range of scenarios from super-generic, to super-specific.

@AndrewIngram thanks so much for the detailed thoughts here, my mind is spinning a bit trying to think of how'd we'd model this in a scalable way and the impact it would have on all of our frontend code. I can definitely see why we'd consider errors as a return type. Any chance you have an example schema that follows this pattern with more examples that I could study?

I'm going to prototype a few ways we could short-term model out some permission utilities for resolvers :)

brianlovin:
@mxstbr here is a proposed solution https://github.com/withspectrum/spectrum/pull/2632/commits/6973607e883605987c1edd6535b0ceb984e22903

It's not perfect yet, so let's iterate and discuss. I feel fairly strongly that we shouldn't have things as simple as `user.isOwner` because that's just barely better than doing `{ isOwner } = loaders.user.load([ userId, channelId])`. But having a function like `canManageChannel` that abstracts away a ton of logic that is used in almost every channel mutation resolver feels like a win. I'm not sure yet how this will expand over time, and I can see why we might not want something as granular as `canCreateChannel` - open to ideas there though!

api/mutations/channel/utils.js
mxstbr: We have to remove this. Loaders are not meant to be used that way and it might lead to security issues down the line.

mxstbr:
Looks great to me. I'm still not a fan of `canCreateChannel` though, that seems to specific to a single mutation to me 😅

brianlovin:
Good call - I should change that to a “canManageCommunith” or something

brianlovin:
Okay those last two commits tidy things up. Now we have permissions:
- `canAdministerChannel` => returns true if user owns the channel, or they are a community owner/mod
- `canModerateChannel` => returns true if user owns or moderates either the channel or community
- `canAdministerCommunity` => return true only if user is owner
- `canModerateCommunity` => returns true if user is community owner or mod

mxstbr:
Looks really good to me!

brianlovin:
cc @orta same issue here with danger, if it's helpful for debugging

orta:
I guess then that it's to do with the size/scope of the PR, given that both of them fail with the same errors. Likely some aggressive pagination for generating diffs, or pulling commit/comment metadata. Maybe they'll just need to chill out a bit.

brianlovin:
Okay @mxstbr, I think this is ready. Would love your final code review and feedback, especially around how I've structured the type system for community and channel settings; tried my best to keep things flexible for future slack event types (new members, new moderation events, etc).

Next steps after approval will be to get this on Alpha and start testing with a few partners.

brianlovin:
Damn. 

Tests were failing, which makes sense given all the new permissions structure and changes to context. I started digging in here and attempted to fix the test suite to use the same extended `user` object we return from `context` - which made me discover another problem.

Turns out the `@isAuthed` directive wasn't working...spent quite a while digging into this, but loosely `graphql-tools` has a new suite of tools for building custom schema directives, and it required a pretty big version bump (1.x => 3.0). But after I upgraded, I can't seem to get the schema directives to work, I just keep getting this error:

`TypeError: visitorClass.getDirectiveDeclaration is not a function`

That happens whenever the API spins up.

I was hoping this would be ready to ship tomorrow, but adding the custom directive is turning out to be a pita and until I fix that I can't fix the failing tests...will work on this tomorrow.

brianlovin:
Um...so I don't even know if I can use this directive style on queries. I'm pretty confused https://www.apollographql.com/docs/graphql-tools/schema-directives.html#What-about-query-directives => this documentation is pretty tricky to understand.

brianlovin:
The more I reflect on this, the more I understand why this solution isn't working. I have some ideas, but it's getting way too late here so will dig in in the morning.

The shortest term fix will be to just rip out the `@isAuthed` directive for now and re-add the redundant checks in resolvers. That will at least get tests to pass and get us back on track.

orta:
Ergh, sorry about Peril here, I made https://github.com/danger/danger-js/issues/580 and https://github.com/danger/danger-js/issues/581 which I hope I can get to tonight

brianlovin:
@mxstbr we lost the rule to ignore console logs in migrations :)

mxstbr:
We never had that rule in the first place, console.logs aren't allowed anywhere—only flow :wink:

/cc @orta that patch with the rate limiting fixed Peril for large PRs! :tada:

brianlovin:
We had an ignore directory rule there. But no worries, will just switch to error

brianlovin:
Writing tests for this is a bit tricky as it would require handling 3rd party oauth (not possibly with cypress afaik) or mocking out entire request flows to and from slack api. I'm honestly not sure how to go about this properly - in the meantime I at least added checks around who should have access to querying slack settings on communities and channels.

So @mxstbr this is ready for your code review. I also refactored the permission logic today to be separated from the context constructor; lmk if you're happy with how all that looks!

mxstbr:
> I also refactored the permission logic today to be separated from the context constructor

Wait hold on, why? I thought we were happy with that?

api/types/ChannelSlackSettings.js
```diff
@@ -0,0 +1,26 @@
...
+  enum BotConnectionEventType {
+    threadCreated
+  }
+  
+  type BotConnection {
+    threadCreated: String
+  }
+
+	type ChannelSlackSettings {
+		botConnection: BotConnection
```
mxstbr: This name is confusing: `botConnection` is not a connection. (as per the Relay spec and how we use it elsewhere) If you want this to be a connection just like all our other connections (which it should be), please make it conformant by adding edges, nodes and page info:

```GraphQL
type BotSettings {
  threadCreated: String
}

type BotConnectionEdge {
  node: BotSettings!
}

type BotConnection {
  pageInfo: PageInfo!
  edges: [BotConnectionEdge!]
}
```

You don't have to build the actual pagination for now, but at least this sets us up to do it properly in the future.

mxstbr:
Getting this locally after connecting a Slack team:

screen shot 2018-05-02 at 10 14 13

Maybe I need some updated secrets or something?

brianlovin:
> Wait hold on, why? I thought we were happy with that?

Asked about this on Twitter and you said it looked good! Anyways, the reason is to keep permissions away from the schema and pushed down into business logic. Writing them as resolver wrappers is still a bit more tightly integrated, but it makes for a nice experience.

> This name is confusing: botConnection is not a connection. 

I don't want it to be a relay connection spec field. I am trying to communicate how the bot is connected to the slack team, so connection came to mind. Open to other naming ideas!

> Getting this locally after connecting a Slack team:

Console logs?

mxstbr:
> Anyways, the reason is to keep permissions away from the schema and pushed down into business logic.

They weren't in the schema before though, just packed into the GraphQL context so that they're easy to call. `@isAuthed` was in the schema, and I thought that's what we were talking about moving out. Anyways, not a biggie, I liked the `user.canManageChannel(channelId)` API more than `canManageChannel(user.id, channelId, loaders)` one but it doesn't matter too much.

> I am trying to communicate how the bot is connected to the slack team, so connection came to mind.

Oh I see, I totally misinterpreted that name. I think we probably shouldn't call it `*Connection` since that's pretty reserved for the Relay stuff—what about `slackSettings.links` or just `slackSettings.botConfig` or something like that?

brianlovin:
Re the error you're seeing after connecting a slack team: for some reason we have 2 spectrum apps on Slack, and the secrets were mismatched. I updated the `now-secrets` in 1password, can you grab that and try again? Any reason we have 2 apps for this?

brianlovin:
> Oh I see, I totally misinterpreted that name. I think we probably shouldn't call it *Connection since that's pretty reserved for the Relay stuff—what about slackSettings.links or just slackSettings.botConfig or something like that?

Will update now.

brianlovin:
Named `botLinks` - how does that feel to you?

PR #2674, opened 2018-03-30
Fix e2e tests

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

@brianlovin we need to run resetdb both before() and beforeEach(),
because otherwise the cy.visit('/whatever') runs before the beforeEach()
so it'd 404.


mxstbr:
Hmm, I made the API auto-restart on crash (by using Nodemon) yet the tests still fail 🤔 Something else must be going on here. Going to dig in and profile now.

mxstbr:
Just merged my CircleCI refactor try from #2742 which seems to have passed on CI—if it doesn't work here I'll remove that change again.

mxstbr:
Only a single test failed in that last run oh my god. 🤞

mxstbr:
So close, and that one failure is an actual failure. (fixed in the last commit) Let's hope that this is it! 🤞 

screen shot 2018-04-10 at 12 54 48

mxstbr:
Argh now they're failing again :cry: I thought for just a second I'd fixed it.

brianlovin:
Unable to figure out the proper way to patch in a check for the existence of the testing db, but I think #2810 is probably a better place for that patch to come in. Since this PR fixes a bunch of tests and frontend code, let's ship it.

PR #2688, opened 2018-04-03
[A11Y] - Add empty alt attribute to homepage images

AgtLucas:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Hi all!

Thanks for making Spectrum open source! :)

This is a small PR, I'm only adding empty alt attribute to the images on the homepage, as I understand them as presentation purpose only. https://www.w3.org/TR/html-aam-1.0/#el-img-emptyalt

Yes, we could use `role="presentation"` instead but the support is not that great.

### DIFF
--- a/src/components/illustrations/index.js
+++ b/src/components/illustrations/index.js
@@ -108,7 +108,7 @@ const DiscoverImage = styled.img`
   }
 `;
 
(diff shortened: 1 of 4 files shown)


mxstbr (approved):
Thank you! :100:

PR #2691, opened 2018-04-03
Fix Tooltip "arrow" bottom position

idanwe:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Really like your tooltip behavior and style, so I played with it and found that.

the current behavior:
screen shot 2018-04-03 at 13 57 14

### DIFF
--- a/src/components/globals/index.js
+++ b/src/components/globals/index.js
@@ -442,8 +442,8 @@ const returnTooltip = props => {
           top: calc(100% + 3px);
           left: 50%;
           transform: translateX(-50%);
-          border-bottom-width: 0;
-          border-top-color: ${
+          border-top-width: 0;
+          border-bottom-color: ${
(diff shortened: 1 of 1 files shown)


mxstbr (changes requested):
Thank you for contributing! 💜 

Interestingly, these tooltips display above the element for me:

screen shot 2018-04-03 at 13 12 35

Which browser are you using? Maybe the better fix would be to make sure the tooltips display above the text always?

idanwe:
Your right, every thing their is ok.

I just changed the props position to bottom in order to demonstrate it is broken in that case.

mxstbr (approved):
Oh, I should've looked at the full context of the code, sorry! 🤦‍♂️ 

This LGTM, thank you so much for contributing!

PR #2728, opened 2018-04-05
Fix invalid token response when using accessToken authentication

paulxuca:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

This change is required because without it, the cookies are outdated on subsequent authentication attempts. An example is while being logged in, and then going to `/auth/PROVIDER?r=https://spectrum.chat/home&authType=token`, `req.cookies` will be outdated until `res.redirect(...)` is called, due to `cookie-session` updating cookies on header change (uses `on-headers` under the hood). By using the `getCookies` function written, it'll always be up-to-date. 

Please let me know if I should provide any more details as to why this is required! :)

### DIFF
--- a/api/routes/auth/create-signin-routes.js
+++ b/api/routes/auth/create-signin-routes.js
@@ -10,7 +10,7 @@
 import passport from 'passport';
 import { URL } from 'url';
 import isSpectrumUrl from '../../utils/is-spectrum-url';
-import { signCookie } from 'shared/cookie-utils';
+import { signCookie, getCookies } from 'shared/cookie-utils';
 
 const IS_PROD = process.env.NODE_ENV === 'production';
 const FALLBACK_URL = IS_PROD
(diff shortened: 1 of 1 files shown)


mxstbr (commented):
That makes sense to me overall (also I'm impressed you managed to find this undocumented functionality :wink:), is there a chance that `req.session.passport` is undefined there and should we handle that case?

paulxuca:
We're also working on something similar with a RN app, so I looked around and found this :) `req.session.passport` being undefined would be a separate issue, but I haven't seen that yet, just `req.cookies` being outdated with data from the previous authentication attempt (i.e Logging in using facebook while being logged in via twitter, and receiving a outdated token).

mxstbr (approved):
Makes sense, this LGTM—let's see if anything bad happens with this code! 🤞

PR #2742, opened 2018-04-05
Split e2e tests from unit tests on CircleCI

mxstbr:
**Status**
- [ ] WIP
- [ ] Ready for review
- [x] Needs testing

Since e2e tests are very flakey right now, this should make it more obvious if CI has passed apart from e2e tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,8 @@ jobs:
   build_web:
     <<: *js_defaults
     steps:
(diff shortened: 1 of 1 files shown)


mxstbr:
Ah damn, I forgot that our unit tests also need the db to run. Hmmm

mxstbr:
Actually, this won't make anything better. Closing.

PR #2752, opened 2018-04-05
Improve NavBar’s a11y

Tom-Bonnike:
# Focus styles:
kapture 2018-04-05 at 22 57 51

# Skip link “in action”:
kapture 2018-04-05 at 23 30 01

# PR Description

Heyo. Here are the list of changes this PR introduce (you can browse the PR commit by commit, that could help you make sense of it):
- Add focus styles to the NavBar 4edab0d71803cd6ca98e2ca7db9cce0190b52862.
- Remove the ability to focus the logo (it links to the same page as the next item in the navbar and it’s just a meaningless icon anyway. It’s still clickable.) edae63007b0c6af31fe95bba301dc31a87543e94.
- Add a skip link 2bb3b8e1c7917460529ebe2f42e93fafaa658c96. [Why you might want that](https://a11yproject.com/posts/skip-nav-links/).
- Add `aria-current="page"` to the link of the current page c485a9632888ad24d389f5e0419ef8f9557732b7. [What it communicates](https://tink.uk/using-the-aria-current-attribute/).
- Filter out invalid props from `<Link>` component a5d5f54571c7d52f33cf2e0a91a34f512cb7bbca. This PR added warnings for the `isHidden` prop but there was already a warning for the `loggedOut` prop on the unauthenticated header. This commit fixes that.

I didn’t touch the `ProfileDropdown`.

**Status**
- [ ] WIP
- [ ] Ready for review
- [x] Needs testing

- hyperion (frontend)

- Add focus styles to the nav bar
- Add skip link to the nav bar
- Improve overall accessibility of the navbar

### DIFF
--- a/src/components/link/props.js
+++ b/src/components/link/props.js
@@ -1,0 +1,49 @@
+// @flow
+// This file exports an array of all props that are valid on an `a` (anchor)
+// element. We need to do that so that we don’t spread invalid props on React
+// Router’s `<Link>` component, which would trigger an error.
+
+const BASE_PROPS = [
+  'id',
+  'className',
+  'tabIndex',
+  'lang',
+  'aria-hidden',
+  'children',
+];
+const FOCUS_EVENTS = ['onFocus', 'onBlur'];
+const MOUSE_EVENTS = [
+  'onClick',
+  'onContextMenu',
+  'onDoubleClick',
+  'onDrag',
+  'onDragEnd',
+  'onDragEnter',
+  'onDragExit',
+  'onDragLeave',
+  'onDragOver',
+  'onDragStart',
+  'onDrop',
+  'onMouseDown',
+  'onMouseEnter',
+  'onMouseLeave',
+  'onMouseMove',
+  'onMouseOut',
+  'onMouseOver',
+  'onMouseUp',
+];
+
+const LINK_PROPS = [
+  'href',
+  'hrefLang',
+  'rel',
+  'target',
+  'autoFocus',
+  'aria-label',
+  'aria-current',
+  'to',
+  'download',
+  'title',
+];
+
+export default [...BASE_PROPS, ...LINK_PROPS, ...FOCUS_EVENTS, ...MOUSE_EVENTS];
--- a/src/views/navbar/index.js
+++ b/src/views/navbar/index.js
@@ -88,6 +88,13 @@ class Navbar extends React.Component<Props, State> {
   handleSkipLinkFocus = () => this.setState({ isSkipLinkFocused: true });
   handleSkipLinkBlur = () => this.setState({ isSkipLinkFocused: false });
 
+  getTabProps(isActive: boolean) {
+    return {
+      'data-active': isActive,
+      'aria-current': isActive ? 'page' : undefined,
+    };
+  }
+
   render() {
     const { history, match, currentUser, notificationCounts } = this.props;
 
@@ -168,7 +175,10 @@ class Navbar extends React.Component<Props, State> {
             Skip to content
           </SkipLink>
 
-          <HomeTab data-active={match.url === '/' && match.isExact} to="/">
+          <HomeTab
+            {...this.getTabProps(match.url === '/' && match.isExact)}
+            to="/"
+          >
             <Icon glyph="home" />
             <Label>Home</Label>
           </HomeTab>
@@ -178,7 +188,7 @@ class Navbar extends React.Component<Props, State> {
           />
 
           <ExploreTab
-            data-active={history.location.pathname === '/explore'}
+            {...this.getTabProps(history.location.pathname === '/explore')}
             to="/explore"
           >
             <Icon glyph="explore" />
@@ -194,9 +204,9 @@ class Navbar extends React.Component<Props, State> {
           <ProfileDrop>
             <Tab
               className={'hideOnMobile'}
-              data-active={
+              {...this.getTabProps(
                 history.location.pathname === `/users/${loggedInUser.username}`
-              }
+              )}
               to={
                 loggedInUser.username ? `/users/${loggedInUser.username}` : '/'
               }
@@ -212,9 +222,9 @@ class Navbar extends React.Component<Props, State> {
 
           <ProfileTab
             className={'hideOnDesktop'}
-            data-active={
+            {...this.getTabProps(
               history.location.pathname === `/users/${loggedInUser.username}`
-            }
+            )}
             to={loggedInUser.username ? `/users/${loggedInUser.username}` : '/'}
           >
             <Icon glyph="profile" />
@@ -246,14 +256,14 @@ class Navbar extends React.Component<Props, State> {
 
           <HomeTab
             className={'hideOnDesktop'}
-            data-active={match.url === '/' && match.isExact}
+            {...this.getTabProps(match.url === '/' && match.isExact)}
             to="/"
           >
             <Icon glyph="logo" />
             <Label>About</Label>
           </HomeTab>
           <ExploreTab
-            data-active={history.location.pathname === '/explore'}
+            {...this.getTabProps(history.location.pathname === '/explore')}
             to="/explore"
             loggedOut={!loggedInUser}
           >
@@ -261,14 +271,14 @@ class Navbar extends React.Component<Props, State> {
             <Label>Explore</Label>
           </ExploreTab>
           <SupportTab
-            data-active={history.location.pathname === '/support'}
+            {...this.getTabProps(history.location.pathname === '/support')}
             to="/support"
           >
             <Icon glyph="like" />
             <Label>Support</Label>
           </SupportTab>
           <PricingTab
-            data-active={history.location.pathname === '/pricing'}
+            {...this.getTabProps(history.location.pathname === '/pricing')}
             to="/pricing"
           >
             <Icon glyph="payment" />
--- a/src/views/navbar/index.js
+++ b/src/views/navbar/index.js
@@ -135,7 +135,7 @@ class Navbar extends React.Component<Props> {
             )}
           </Head>
 
-          <Logo to="/">
+          <Logo to="/" aria-hidden tabIndex="-1">
             <Icon glyph="logo" size={28} />
           </Logo>
 
@@ -198,7 +198,7 @@ class Navbar extends React.Component<Props> {
     if (!loggedInUser) {
       return (
         <Nav hideOnMobile={hideNavOnMobile} loggedOut={!loggedInUser}>
-          <Logo to="/">
+          <Logo to="/" aria-hidden tabIndex="-1">
             <Icon glyph="logo" size={28} />
           </Logo>
           <HomeTab
--- a/src/views/navbar/index.js
+++ b/src/views/navbar/index.js
@@ -20,6 +20,7 @@ import {
   Tab,
   Label,
   Navatar,
+  SkipLink,
 } from './style';
 
 type Props = {
@@ -36,9 +37,21 @@ type Props = {
   activeInboxThread: ?string,
 };
 
-class Navbar extends React.Component<Props> {
-  shouldComponentUpdate(nextProps) {
+type State = {
+  isSkipLinkFocused: boolean,
+};
+
+class Navbar extends React.Component<Props, State> {
+  state = {
+    isSkipLinkFocused: false,
+  };
+
+  shouldComponentUpdate(nextProps, nextState) {
     const currProps = this.props;
+
+    // If the update was caused by the focus on the skip link
+    if (nextState.isSkipLinkFocused !== this.state.isSkipLinkFocused)
+      return true;
 
     // if route changes
     if (currProps.location.pathname !== nextProps.location.pathname)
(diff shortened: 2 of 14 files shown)


AgtLucas:
> Removes the ability to focus the logo (it links to the same as the next item in the navbar and there’s just a meaningless icon anyway)

Nice catch!

src/views/navbar/style.js
```diff
@@ -346,3 +384,17 @@ export const MarkAllSeen = styled.span`
       props.isActive ? props.theme.brand.default : props.theme.text.alt};
   }
 `;
+
+// We make it a real link element because anchor links don’t work properly with React Router.
+// Ref: https://github.com/ReactTraining/react-router/issues/394.
+export const SkipLink = Tab.withComponent('a').extend`
```
Tom-Bonnike, line 356: I didn’t find any occurence of `.extend` (or `withComponent(…).extend`) in your codebase, so I assume you might not like this for whatever reason? 😄

src/components/appViewWrapper/index.js
```diff
@@ -7,11 +7,16 @@ import { withRouter } from 'react-router';
 import { Wrapper } from './style';
 
 const AppViewWrapperPure = (props: Object): React$Element<any> => (
-  // Note(@mxstbr): This ID is needed to make infinite scrolling work
-  // DO NOT REMOVE IT
-  <Wrapper {...props} id="scroller-for-thread-feed">
-    {props.children}
-  </Wrapper>
+  <main id="main">
```
Tom-Bonnike: This doesn’t mess up the styles and there aren’t any `<main>` in your codebase so I guess it’s fine. I also added/replaced this `<main>` tag in other places (could be done in more).

Tom-Bonnike (commented):
There might be more spots where we need to add the `id="main"` if you want to keep the skip link.
I think I tried most pages when logged in. I probably need to go through more pages and find the main container, especially when not authenticated.

src/views/navbar/style.js
```diff
@@ -197,6 +199,16 @@ export const Logo = styled(Tab)`
...
+
+  ${props =>
+    props.isHidden &&
+    css`
+      position: absolute;
+      width: 1px;
+      height: 1px;
+      overflow: hidden;
+      clip: rect(1px, 1px, 1px, 1px);
+    `};
```
Tom-Bonnike: I guess we could re-use the `VisuallyHidden` component styles instead? Or simply use `display: none` as there’s an `aria-hidden` on the logo anyway.

AgtLucas: ~~@Tom-Bonnike But this is the logo, right? So it needs to be shown visually, hence we cannot use `display: none`, we just need to prevent the focus (as you pointed), so using `aria-hidden` is fine.~~

Edited by @AgtLucas due https://github.com/withspectrum/spectrum/pull/2752/files#r179607595 😅

Tom-Bonnike: Yah, but these styles are here to hide it (from all users), whenever the skip link is focused. So `display: none` would probably be fine. :)

AgtLucas: Ah, I haven't notice the `props.isHidden` prop :)

src/views/navbar/index.js
```diff
@@ -135,21 +151,47 @@ class Navbar extends React.Component<Props> {
             )}
           </Head>
 
-          <Logo to="/">
+          <Logo
+            to="/"
+            aria-hidden
+            tabIndex="-1"
+            isHidden={this.state.isSkipLinkFocused}
```
Tom-Bonnike, line 165: Apparently, you don’t whitelist style props that go on `<Link>` components, so they are added to the DOM element and throw a React warning… It also happens for the `loggedOut` prop in the logged out header. I didn’t try to fix it. I guess it’s just a matter of filtering them out in `src/components/link`?

src/views/navbar/style.js
mxstbr: We've been using styled-components since before that was a thing, LGTM!

src/views/navbar/index.js
mxstbr: Yeah should just be a matter of filtering them—those console logs have been annoying me for a long time so a fix for this would be amazing!

mxstbr:
This looks amazing! 😍 Can't wait to ship this when it's done! Do you want to continue working on this branch or should we start shipping incremental improvements?

Tom-Bonnike:
Apart from minor changes, I’m pretty much done with this PR. Improving the `ProfileDropdown` and its links probably deserves its own PR. I didn’t touch any of it at all.

About the skip link: because you don’t seem to have a shared wrapper/container/layout component for every page, I need to add the `id="main"` in multiple components and, again, it’s probably missing from some pages. I know for a fact it’s missing on almost every unauthenticated page.
Do you think you could go through and add it to the rest of the pages yourself? I’m not sure from which pages it’s missing, and then I would need to find the component by looking at the React devtools… which isn’t so easy 😄

Otherwise I’ll try to find some time this week-end.
It’s not so bad if it’s missing from a page anyway. I’d recommend that we use the `<main>` tag if it’s not used elsewhere on the page, too. 🤷‍♂️

src/views/navbar/index.js
```diff
@@ -198,19 +246,39 @@ class Navbar extends React.Component<Props> {
...
+            onBlur={this.handleSkipLinkBlur}
+          >
+            Skip to content
+          </SkipLink>
+
           <HomeTab
             className={'hideOnDesktop'}
             data-active={match.url === '/' && match.isExact}
+            aria-current={
+              match.url === '/' && match.isExact ? 'page' : undefined
```
Tom-Bonnike: I should probably extract all of those to a method that would return the `data-active` + `aria-current` props, as they use the same condition.

src/views/navbar/index.js
Tom-Bonnike: ~I think https://github.com/withspectrum/spectrum/pull/2752/commits/a2645419d9f9e1f6a0e96f35cf16e65766dbd6f1 is the way to go.
I might be missing some of your custom `data-…` attributes (`data-cy`, maybe?), but that’s the only foolproof solution I know of.~

Tom-Bonnike: Oh. I’m realizing now I could probably whitelist all props starting with `data-` as they’re all valid 🤔Should I do that, @mxstbr?

EDIT: done with a5d5f54571c7d52f33cf2e0a91a34f512cb7bbca.

Tom-Bonnike:
OK. I went through most pages that have the nav bar and it seems like I didn’t actually forget any, except for the “single thread” view. If I try to add the `main` id to the `Thread` component, I end up with a duplicate id on the home… I’m not sure where to put it 🤷‍♂️ 

I’m wondering if we couldn’t simply add a “global” `<main id="main> … </main>"` in `routes.js` that would wrap the `<Switch>` in there, instead of having to add the id manually on every page/view.
Whaddya think?

mxstbr:
> I’m wondering if we couldn’t simply add a “global” <main id="main> … </main>" in routes.js that would wrap the <Switch> in there, instead of having to add the id manually on every page/view.
Whaddya think?

That sounds very reasonable to me, let's do that instead!

Tom-Bonnike:
OK, that works fine, and it‘s much more clean like this.
There’s a small page jump on the first focus after using the Skip to content button now, I’m not too sure why. It wasn‘t happening before, but that’s not a big issue.

Ready for review 🎉

src/views/navbar/style.js
```diff
@@ -346,3 +350,17 @@ export const MarkAllSeen = styled.span`
       props.isActive ? props.theme.brand.default : props.theme.text.alt};
   }
 `;
+
+// We make it a real link element because anchor links don’t work properly with React Router.
+// Ref: https://github.com/ReactTraining/react-router/issues/394.
+export const SkipLink = Tab.withComponent('a').extend`
+  grid-area: logo;
```
AgtLucas, line 357: Sorry for this question, but I'm not familiar with CSS grid, why this `logo` here?

src/components/link/props.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+// This file exports an array of all props that are valid on an `a` (anchor)
```
AgtLucas, line 2: Don't we need to whitelist `role` as well?

src/views/navbar/style.js
Tom-Bonnike: The areas of the grid are defined here:
https://github.com/Tom-Bonnike/spectrum/blob/eef68777727b8bda4c0bf908a8ef41e01fe203c9/src/views/navbar/style.js#L10

It’s just a way to slice your grids into named areas.

Because the skip link replaces the logo, I simply re-use that one.

src/components/link/props.js
Tom-Bonnike: Tbh, I copied that from another codebase I work on. It’s probably missing a lot of attributes, including `role`, indeed. Probably a large number of `aria-…` attributes too. 

IMO this is the most foolproof solution, but it would be much better if a module handled that for us, maintaining it is a pain.

Two modules exist but aren’t so popular:
- https://www.npmjs.com/package/filter-react-props (it seems to have a pretty complete list: https://github.com/asistapl/filter-react-props/blob/master/src/index.js)
- https://badge.fury.io/js/filter-react-dom-props 

I wish this was exposed by `react-dom` instead!

Tom-Bonnike: But yeah, I should probably add `role` to the list. Ideally you’d add new attributes whenever you’re using one that isn’t whitelisted, but that’s annoying.

On second thought, I think we should simply filter out `isHidden` and `loggedOut` like it was done for the `react-router` props, and leave a comment explaining what they are, something like this:
```js
export default withRouter({
  // Filter the `withRouter` props from being passed through to <Link />
  staticContext,
  history,
  location,
  match,

  // Styling props
  isHidden,
  loggedOut,

  ...rest
}: Props) => (
  <Link
    {...rest}
    onClick={…}
  />
))
```

That still require us to maintain a list, though.

src/views/navbar/style.js
AgtLucas: Ah, I see! Thanks!

AgtLucas:
LGTM!

src/components/link/props.js
mxstbr: We actually maintain that list for styled-components! For now it should be good enough to specifically filter out those props though, that looks fine to me.

Tom-Bonnike: So, you’re fine with how it is right now? Or should I change it to what I proposed in my previous comment?

brianlovin:
<3 <3 to this!

Clicking around:
- I didn't notice an `aria-current` getting set while viewing the messages tab or the notifications tab
- maybe lower pri, or a follow on pr, but should we also handle the signed-out nav? there are two:
  - in-app signed out nav has links to explore/support/pricing
  - marketing pages signed out nav has links to pricing/support/about
- i'm pretty sure adding the `div id='main'` component to wrap our routes breaks our mobile layout - if you resize the browser you'll notice that the navbar gets pushed offscreen bottom. cc @uberbryn as maybe there's a quick fix for that?

brianlovin (changes requested):
Main thing is that the `div id="main"` seems to be breaking our mobile layout and pushing the nav offscreen bottom.

Tom-Bonnike:
- Oh oops I totally forgot to test these those two but it makes sense, I’ll fix that 😅
- Yeah, I didn’t touch these navs. Could be a follow on PR.
- I didn’t notice that. I’ll check again. Maybe a `flex` issue.

Thank you.

Tom-Bonnike:
OK, I rebased and fixed the `aria-current` not being added on the messages and notifications tab.

The `<main>` breaking the layout is indeed a flex issue because of the additional wrapper.

screen shot 2018-04-08 at 11 10 02

This fixes it (a simple `display: flex; flex: 1 1 100%;` might even work):
```js
const Main = Body.withComponent('main')

<Main id="main"> … </Main>
```

The issue is that the first component inside the `main` has a `max-height: calc(100% - 48px)`, because it needed to take into account the size of the header. This isn’t needed anymore and breaks the layout on some pages. The messages tab for instance.

screen shot 2018-04-08 at 11 09 37

I’m not sure where it would still be needed and if I should remove it everywhere, though… :< There might be other layout issues on other pages but I’m not sure if I’d even notice them :/

brianlovin:
@Tom-Bonnike sorry for slow reply - I tested both of your proposed fixes on your branch and I think both of those will do the trick:
- add flex to the main element
- change all instances of height being `calc(100% - 48px)` to `100%`

If you're up for making those quick swaps, we can get this on alpha and test for a few days before shipping to production :)

Tom-Bonnike:
@brianlovin Hehe, no worries. I was just not feeling confident changing this layout without first having your green light.
https://github.com/withspectrum/spectrum/pull/2752/commits/25428b825bade8c28c2f63fd29e8f9f68d01e409 should do it.

brianlovin:
Absolutely awesome! Clicking around to everything I can think of locally on mobile + desktop viewports and this is working great. I'm going to merge and ship this to `alpha.spectrum.chat` first for further testing over the next couple days.

🎉🎉

brianlovin (approved):

brianlovin:
Interestingly after I merged this into alpha and pulled the changes, views started to break with the missing calc statement. Digging in.

brianlovin:
@uberbryn has some ideas on how to reduce the complexity of our app layout, which has become a bit of a maintenance nightmare :)

PR #2778, opened 2018-04-07
Paste image from clipboard into chat input

StefanNemeth:
**Status**
- [ ] WIP
- [x] Ready for review
- [x] Needs testing

- hyperion (frontend)

This change enables users to paste images directly into the chat input (Ctrl + V) as requested in https://github.com/withspectrum/spectrum/issues/2755. The image is previewed before it can be posted with an optional message as shown in the GIF below.

kapture 2018-04-07 at 15 16 10

### DIFF
--- a/src/components/chatInput/components/mediaUploader.js
+++ b/src/components/chatInput/components/mediaUploader.js
@@ -123,7 +123,7 @@ class MediaUploader extends React.Component<Props> {
         <MediaLabel>
           <MediaInput
             type="file"
-            accept={'.png, .jpg, .jpeg, .gif, .mp4'}
+            accept={'.png, .jpg, .jpeg, .gif'}
             multiple={false}
             onChange={this.onChange}
           />
--- a/src/components/chatInput/components/mediaUploader.js
+++ b/src/components/chatInput/components/mediaUploader.js
@@ -11,34 +11,31 @@ import {
 } from 'src/helpers/images';
 
 type Props = {
-  onValidated: Function,
+  onValidatedUpload: Function,
+  onValidatedPreview: Function,
   onError: Function,
   currentUser: ?Object,
   isSendingMediaMessage: boolean,
+  isFocused: boolean,
 };
 
 class MediaUploader extends React.Component<Props> {
   form: any;
 
-  validateUpload = (validity: Object, file: ?Object) => {
+  validate = (validity: Object, file: ?Object) => {
     const { currentUser } = this.props;
 
-    if (!currentUser)
-      return this.props.onError('You must be signed in to upload images');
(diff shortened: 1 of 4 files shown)


mxstbr (commented):
Amazing, thank you so much for contributing!

I'm not 100% across the uploading code, I'll defer to @brianlovin for that, I'm mainly wondering whether we can make the same UX happen for normal, non-pasted media messages too? That'd be great so it's consistent no matter how you add the image.

StefanNemeth:
Thanks for the quick feedback! Non-pasted images get a preview now as well.

I wonder whether MP4 and other non-image formats are accepted because I cannot see any "image" validation in the API code (maybe I'm just missing something)?

mxstbr:
We should definitely be validating that—ref #2494 I think somebody's working on a backend PR for it?

brianlovin:
@StefanNemeth this is <3 <3 - super excited to get this in!

One request:
- could we add some sort of ui to the image preview, like an 'x' on the top right corner or something, that allows the user to click or tap to remove the image preview? i see that it's possible to type backspace to clear it, but it'd be great to have some ui for this :)

brianlovin:
Oh, and the bottom border radius of the image preview is pretty weird. Going to upload a before/after proposal :)

brianlovin:
I made a super rough figma mock of how this might be cleaned up: https://www.figma.com/file/Ul0ngM7Janfdyn8DAdOtPYhB/Image-preview

Items to fix:
- add ui for removing image preview
- fix border radius on image preview
- consistent left/top padding on image preview

And as Max pointed out, let's go ahead and validate file types on the client as well, and we'll make sure it's server side validated with #2494 :)

brianlovin (changes requested):
See figma link above

StefanNemeth:
@brianlovin Just added the UI you proposed, looks way better now 👍 

bildschirmfoto 2018-04-08 um 18 02 50

brianlovin:
@StefanNemeth what was the reason for removing mp4 support in that last commit? Just want to make sure we're staying aligned with our backend as well on supported file types :)

StefanNemeth:
@brianlovin Oh, I thought that MP4 isn't supposed to be accepted because it says "upload photo", oopsie!

brianlovin:
> Oh, I thought that MP4 isn't supposed to be accepted because it says "upload photo", oopsie!

Totally reasonable! Our copy there isn't ideal...also I don't think I've seen one person upload an mp4 as a profile/cover photo...lets allow it for now and constrain down as needed :)

brianlovin:
I'm going to get this on alpha.spectrum.chat here in a sec :)

brianlovin (approved):

PR #2780, opened 2018-04-07
Better code handling in messages

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- Much improved code support in messages, just type ``` and you're ready to go!

On top of the obvious experience changes this also makes it so our API doesn't accept invalid DraftJS state input, which should help keep the server from crashing.


mxstbr:
This is getting pretty close, mostly needs some styling work. @brianlovin mind doing a quick cleanup on the styles of this?

mxstbr:
Demo:



(ignore the bug where you don't enter the code block when it's created, working on that. ref https://github.com/withspectrum/draft-js-markdown-plugin/issues/41)

The last behavioral things to figure out:

- [x] How to exit code blocks, this is broken somehow even though it works from the plugin 🤔 
- [ ] Rendering syntax highlighting

mxstbr:
The code editor plugin is breaking escaping code blocks with ```, going to dig into that and maybe just disable it temporarily.

mxstbr:
Just an ordering issue, fixed in that last commit. Apart from syntax highlighting (which will be a bit harder to do, so shouldn't be a blocker) this is ready from a behavioral standpoint, just needs styling tweaks!

mxstbr:
@brianlovin since this has quite a few changes to the chat input, I'll branch off of here to start tackling Android support.

PR #2781, opened 2018-04-07
adds membership lists to the left profile column on mid-size displays

superbryntendo:
Also, refactors granularProfile frontend code to be a lot nicer. I have a little more work to do here, but I'm waiting on a download of the latest backups to test with prod data.

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- adds membership lists to the left profile column on mid-size displays

### DIFF
--- a/src/components/granularUserProfile/index.js
+++ b/src/components/granularUserProfile/index.js
@@ -31,6 +31,7 @@ type Props = {
   isCurrentUser?: boolean,
   reputation?: number,
   messageButton?: boolean,
+  multiAction?: boolean,
   children?: React.Node,
   onlineSize?: 'small' | 'large',
   history: Object,
@@ -57,6 +58,14 @@ class GranularUserProfile extends React.Component<Props> {
     this.props.history.push('/messages/new');
   };
 
+  componentDidMount() {
+    if (React.Children.count(this.props.children) > 1) {
+      this.setState({
+        tooManyActions: true,
+      });
+    }
+  }
+
   render() {
     const {
       userObject,
@@ -69,11 +78,12 @@ class GranularUserProfile extends React.Component<Props> {
       badges,
       children,
       messageButton,
+      multiAction,
       onlineSize,
     } = this.props;
 
     return (
-      <Row avatarSize={avatarSize}>
+      <Row avatarSize={avatarSize} multiAction={multiAction}>
         {profilePhoto && (
           <Avatar
             src={profilePhoto}
@@ -99,21 +109,16 @@ class GranularUserProfile extends React.Component<Props> {
           )}
         </LinkHandler>
         {description && <Description>{description}</Description>}
-
-        <Actions>
-          {messageButton &&
-            !children && (
-              <MessageIcon
-                tipText={name ? `Message ${name}` : 'Message'}
-                tipLocation={'top-left'}
-                onClick={this.initMessage}
-              >
-                <Icon glyph="message-new" size={32} />
-              </MessageIcon>
-            )}
-
-          {children}
-        </Actions>
+        {messageButton && (
+          <MessageIcon
+            tipText={name ? `Message ${name}` : 'Message'}
+            tipLocation={'top-left'}
+            onClick={this.initMessage}
+          >
+            <Icon glyph="message-new" size={32} />
+          </MessageIcon>
+        )}
+        <Actions>{children}</Actions>
       </Row>
     );
   }
--- a/src/components/granularUserProfile/index.js
+++ b/src/components/granularUserProfile/index.js
@@ -38,6 +38,7 @@ type Props = {
   isCurrentUser?: boolean,
   reputation?: number,
   messageButton?: boolean,
+  withDescription?: boolean,
   children?: React.Node,
   isOnline?: boolean,
   onlineSize?: 'small' | 'large',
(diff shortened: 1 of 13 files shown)


mxstbr (dismissed):
This LGTM code-wise!

superbryntendo:
👍 I think it's close but pls don't merge yet.

mxstbr:
Sure, let's let @brianlovin give this a review too I guess

superbryntendo:
This is good to go now. Doesn't really change much styling, just makes it usable at more sizes and makes the code a bit cleaner.

brianlovin:
A few notes:
1. Flow is failing ;) Lmk if I can help with that
2. If you have the apollo dev tools installed, open a community profile and view the list of queries run:
screenshot 2018-04-07 19 00 40

You'll notice that the `getCommunityChannels` and `getCommunityMembers` queries ran twice. The reason that this is happening is because the `MetaMemberships` component is being rendered but set to `display: none` - the css there only hides the dom nodes, but does not stop the nodes from executing their queries. Now, this isn't totally bad because we're using https://www.apollographql.com/docs/react/advanced/network-layer.html#query-deduplication to de-duplicate queries (you can confirm this by looking in the network tab of devtools to see only one outbound api request for each of those queries). However, it'd be preferable to just solve this so there's only one set of queries + fewer chances of apollo networking + caching issues.

Maybe a couple approaches to fix:
- use `grid` to reorganize components at that mid-sized viewport so that there is only ever one component for channels and one component for members in the dom
- use some kind of render function that gets called after `componentDidMount` which will have access to `window.innerWidth` to determine if the viewport size and what to render. Generally not a fan of this as you'd probably need some sort of state setting, which complicates things.

Is it possible to fix that using only `grid`?

3. Should the order of the modules be preserved on that mid sized screen? E.g. team members on top, then channel list?
4. The changes to `granularUserProfile` are breaking - I had the `username` and `userObject` props specifically to handle two scenarios:
- prop `username` tells component to render the username in the granular profile
- prop `userObject.username` provides raw data to handle the `LinkHandler` function
I also think the `withDescription` breaks the naming scheme for this component - if you want it to use that pattern then every prop should be `withProfilePhoto`, `withReputation` etc etc. I think this is not ideal because `withFoo` has the semantics that would expect a boolean value. But there are cases where we'd have to add two props like `withReputation={true} reputation={user.reputation}` which is messier than just checking for the presence of `props.reputation` to determine if the developer wants reputation to render in their instance of the profile.

Let me know if that makes sense. I'm totally down for the style changes and moving to be grid powered (although please be sure to check as many possible cases of prop combinations on desktop+mobile to ensure this works, and double check things like the `PendingUsers` component in a private channels settings to make sure the block, approve, and message actions flow properly in the dom).

5. Re this change: https://github.com/withspectrum/spectrum/pull/2781/commits/8dc14d39bdc42869a59463c3eb30004d5e6c2b87 - can you also double check the channel members component (both on channel profile and in public + private channel settings) to see if it's needed/unneeded there? 
6. On this commit, can this just switch to use a `styled-component`? Or is the `className` the best thing to make this work? https://github.com/withspectrum/spectrum/pull/2781/commits/a1f150f35bea7181590ac93f813dd03d0be366b2. Don't mind either way, but we don't use `className` too much unless it's a super hacky workaround for something
7. Small last thing that I always forget, but if you remove a bunch of stuff can you check eslint warnings as well and clean up any unused code? <3

brianlovin (dismissed):
Comments above

superbryntendo:
good feedback!

> use grid to reorganize components at that mid-sized viewport so that there is only ever one component for channels and one component for members in the dom

I think this is the right solution and at first glance, looks probably easy. I'll see what I can come up with.

> Should the order of the modules be preserved on that mid sized screen? E.g. team members on top, then channel list?

So, I think team is important, but not important enough to push channels offscreen. I think it's fine to have team on top on the right, but I think it's most likely that people will care more about their channel memberships when only one of the two can show "above the fold"

> The changes to granularUserProfile are breaking - I had the username and userObject props specifically to handle two scenarios:
prop username tells component to render the username in the granular profile
prop userObject.username provides raw data to handle the LinkHandler function
I also think the withDescription breaks the naming scheme for this component - if you want it to use that pattern then every prop should be withProfilePhoto, withReputation etc etc. I think this is not ideal because withFoo has the semantics that would expect a boolean value. But there are cases where we'd have to add two props like withReputation={true} reputation={user.reputation} which is messier than just checking for the presence of props.reputation to determine if the developer wants reputation to render in their instance of the profile.

I get this now, but it was really unclear then bc it breaks away from all our other patterns. My goal was just to make it match our other patterns and add some booleans for cases that were less likely to be global

> Re this change: 8dc14d3 - can you also double check the channel members component (both on channel profile and in public + private channel settings) to see if it's needed/unneeded there?
On this commit, can this just switch to use a styled-component? Or is the className the best thing to make this work? a1f150f. Don't mind either way, but we don't use className too much unless it's a super hacky workaround for something

This component has been a pain generally. I wish we could handle permissions without adding a `<div>`. But that's not what this is about. Because this component gets used in several unrelated places, I decided to use a classname to set its styling at the location it will be used. The other option would be to make it a styled-component and import it every time we want to adjust it which seems non-ideal.

> Small last thing that I always forget, but if you remove a bunch of stuff can you check eslint warnings as well and clean up any unused code? <3

nah. 😛

brianlovin:
> This component has been a pain generally. I wish we could handle permissions without adding a `<div>`.

One thing you can do is wrap adjacent sibling elements in `<React.Fragment>{...}</React.Fragment>` to avoid creating unnecessary wrapper nodes

superbryntendo:
@brianlovin since that component has an onClick handler, I'm not sure I can replace it with a react.fragment...

brianlovin:
Ah, right - good call, best leave it then.

superbryntendo:
I think this addresses @brian's feedback across the board. Flow didn't report any issues during dev, so we'll see how tests go...

brianlovin:
Small flow fix pushed @uberbryn - just a heads up you'll see the `test_static_js` ci fail if flow fails, or a `yarn run flow` should surface this. No biggie though! Taking a look at everything else now

brianlovin:
@uberbryn will need some wrapping fixes here on mobile:

screenshot 2018-04-10 14 33 47

Places to test granular user profile right now:
- [x] Community profile, team section
- [x] Community profile members list
- [x] Community settings members list
- [x] Public channel settings members list
- [x] Private channel settings members list
- [x] Private channel settings pending users list
- [x] Private channel settings blocked users list
- [x] Channel profile members list

Overall I think this is very close, it just needs better wrapping for a row with multiple actions, like the pending users section with 'approve' 'block' and 'message' as actions.

brianlovin (dismissed):
Needs fixes for granular user profiles with many actions to wrap appropriately on mobile

superbryntendo:
This isn't my favorite solution, but it handles it ok. I focused on wrapping but looking back, I'd rather have dropdowns a la community members' list (though I think this setup is slightly more efficient in batch cases) or ONLY allow list-level batch actions where you check a checkbox next to the users you want to modify and then contextual actions pop up. Not really sure.

I'd prefer to get this shipped and change that later though.

brianlovin (approved):

mxstbr (approved):
LGTM code-wise!

PR #2785, opened 2018-04-08
Clear search input after user switches between communities

LetItRock:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- when user will switch between communities the search input will be cleared;

### DIFF
--- a/src/reducers/dashboardFeed.js
+++ b/src/reducers/dashboardFeed.js
@@ -22,6 +22,10 @@ export default function dashboardFeed(state = initialState, action) {
     case 'SELECT_FEED_CHANNEL':
       return Object.assign({}, state, {
         activeChannel: action.channelId,
+        search: {
+          ...state.search,
+          queryString: '',
+        },
       });
     case 'REMOVE_MOUNTED_THREAD_ID':
       return Object.assign({}, state, {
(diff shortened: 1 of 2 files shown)


brianlovin:
Just curious if we want this cleared when switching communities. A use case where you'd want it persisted is for people who can't quite remember where something was posted.

For example, I'm looking for a thread about redux. I start my search in the React community, but can't find it. I click onto the Frontend Cafe community and the search restarts with my query, but under the new community filter.

This also works well for drilling down from everything => community => channel, where each time we can assume people are seeking more granularity in their search, and persisting the search query will automatically retry the search in their new filter scope.

LetItRock:
Agree on that. 

I can only imagine a situation when you want that search to clear automatically --> is when you were trying to find something about for ex. react and you got it and then you just wanted to check something else in other communities, so you will need to remember to clear this search input before you switch to another community.

Maybe the option here would be to have something like small hold button inside search? :D
screen shot 2018-04-08 at 00 18 54

brianlovin:
Agree on your flow as well and how that behavior could be problematic; given that we do expose a 'clear' button in the search input when a term is there, perhaps that's an acceptable middle ground for the time being?

Basically I would want any changes that ship to have a more compelling user experience in one direction or the other, and right now I can see the pros and cons of both our proposals equally so would defer to just keeping the existing behavior until we can find a meaningful way to improve it.

What do you think?

LetItRock:
Yes, I agree clear button is always there, use it if you need to. 
I think we should not to over-engineer with that search input for now :)

brianlovin:
Thanks @LetItRock - I'll keep thinking on this in the back of my mind though, appreciate you bringing up the issue :)

PR #2793, opened 2018-04-09
Add docker stack.

fubhy:
Don't get scared. This is a work in progress.


admin/src/views/users/components/search/index.js
```diff
@@ -8,7 +8,7 @@ import { withApollo } from 'react-apollo';
 import { withRouter } from 'react-router';
 import { Spinner } from '../../../../components/globals';
 import { throttle } from '../../../../helpers/utils';
-import { searchUsersQuery } from 'shared/graphql/queries/search/searchUsers';
+import { searchUsersQuery } from '@spectrum/shared/graphql/queries/search/searchUsers';
```
FezVrasta: I don't think Flow supports this kind of aliases

fubhy: Yes it does via module.name_mapper

fubhy: Heck they should be using typescript anyways! :D 

Shots fired!!!

docker-compose.yml
```diff
@@ -1,6 +1,20 @@
...
 
 services:
+  redis:
+    image: redis:3.2.11
+    ports:
+      - 6379:6379
+    networks:
+      - spectrum
+
+  rethinkdb:
```
BranLiang, line 34: Volumn connection for data persistence?

fubhy: Yes, definitely. Currently annoying though. Will add later.

brianlovin:
👀 watching this with keen interest! I've never used docker before

Ponjimon:
image
Wtf

But I like Docker, it's nice! @brianlovin you will like it! :tada:

fubhy:
Yo @mxstbr, this is my second attempt at this now. Let's do it step by step... It's an enormous task to get perfectly right. Hence, I guess it's best if we try and make this transformation happen in small, incremental steps. This PR now only dockerizes the services. Nothing else.

Locally, you should be able to run:

```
make bootstrap
make db:up && make db:seed
make start
```

In order to bring up a full range of services (api, et al) in dev mode. The rest (frontend, etc.) is unchanged and should just work.

Before the full potential of this docker stack becomes visible, however, we we have to first unscramble the node modules and properly set up all the package.json of all the workspaces. That means we have to do quite a few things also to the code and how it is currently structured. Once that is done we can also dockerize the other parts and eventually also switch out the entire CI configuration to run with Make and docker-compose.

fubhy:
Deployments with now.sh should also be possible now. We just need to push to dockerhub and then deploy each service's Dockerfile.

docker-compose.yml
```diff
@@ -0,0 +1,187 @@
...
+
+services:
+  redis:
+    image: redis:3.2.11
+    ports:
+      - 6379:6379
+    networks:
+      - spectrum
+
+  rethinkdb:
```
xcaptain, line 34: Why add a service subdirectory, I think it's enough to use /.

fubhy: Because I also added yarn workspaces and this is a preparation step for cleaning up the node modules dependency mess across the board. It's also helpful for keeping the docker context small without maintaining a massive dockerignore file.

mxstbr:
Hmm, I'm getting errors for all services saying "Cannot find file `package.json` in folder `app/<service>`, any clues why?

fubhy:
Not sure! Are you getting that error inside of Docker or when running the services locally?

fubhy:
Did you run "make bootstrap" again after the recent changes?

fubhy:
Try this: 

```
make bootstrap
docker-compose build --no-cache
make start
```

mxstbr:
screen shot 2018-04-13 at 10 57 07

fubhy:
Fixed that.

mxstbr:
Hmm, pulled the latest changes, still getting that `cross-env` error? I've run the three steps from above again

fubhy:
@mxstbr Hmmm. Can't reproduce that: 
screenshot from 2018-04-13 14-35-20

mxstbr:
Oh hold on, do i have to run yarn before make bootstrap?

fubhy:
``make bootstrap`` will produce a docker container that contains the installed node_modules as well as a base image (based on alpine with node.js) that is then used by the individual services that we boostrap with docker-compose. Each service then just copies over the node_modules generated during ``make bootstrap``. That greatly reduces the build time (because we don't have to install the same node modules into each container separately which would be crazy).

So, if you run "make bootstrap" freshly, it should produce a new imager which should contain cross-env because I just re-added that in the last commit.

After that, ``docker-compose build --no-cache api`` will rebuild the api service (if you want to rebuild all, just leave out that argument, but that will take longer... for testing if cross-env works, that's enough).

If it doesn't work, you can ``sh`` into the container using ``docker-compose run --rm api sh``. You should automatically be in ``/app/services/api``. Try running ``yarn run dev`` manually. If that doesn't work, check if ``/app/services/api/node_modules/.bin/cross-env`` is there. If it's not, we have to figure out why. E.g. by running ``cat /app/services/api/package.json || grep cross-env``

mxstbr:
Running `yarn` before `make bootstrap` fixed it—which makes sense given your explanation!

mxstbr:
Played around with this for a bit and it's so much neater than what we currently have! Main question is: can we make the default `make start` only start the minimum necessary services (`api`, `athena` and `mercury`, maybe `hermes`) and then have a CLI flag or something to start the full suite? (you almost never need to run `pluto`, `vulcan` and `chronos` locally, unless you're developing them)

Remaining tasks as far as I can see:

- [ ] Implement CI
- [ ] Publish to DockerHub from CI
- [ ] Update all the docs

Anything else? How can I help with landing this?

**Bonus**

- [ ] Auto-deploy on merge to `alpha`

fubhy:
@mxstbr @brianlovin I did another fresh start without moving any files around. The merge conflicts were still getting out of hand. We shouldn't be seeing those anymore except for yarn.lock file changes.

mxstbr:
Tried testing this just now, but I get this error on startup of every server on `make start`:

screen shot 2018-04-24 at 12 01 01

I've run `yarn` then `make bootstrap` then `docker-composer build --no-cache` then `make start`. Any clues? Some docs would be nice here...

fubhy:
@mxstbr Running "yarn" outside of the container has no effect. The node modules are encapsulated in the container. All you need to do normalle is "make bootstrap && make start". If dependencies have changed "make bootstrap" (builds the base image), then "make build" (rebuilds the individual containers).

But for the normal workflow (daily work), a simple "make start" is enough

fubhy:
I just did a completely fresh setup and can't reproduce your problem. Can you try and check out the PR's branch from scratch?

mxstbr:
Started from scratch, now getting this error on `make bootstrap`:

screen shot 2018-04-25 at 12 44 10

fubhy:
Ah yeah, so the device you store your local docker images on is full. By default it mounts into /var/lib/docker. I presume that you have accumulated quite a few docker images there in the meantime. Especially also because of this PR which caused quite a few rebuilds in the past. You might want to clear your docker images and, if you have set up your system with a separate, larger partition for your home directory, use that for storing the images instead.

https://www.google.com/search?q=docker+change+volume+location

Delete all youir container: ``docker rm $(docker ps -a -q)``
Delete all your images: ``docker rmi $(docker images -q)`` 

Note that this is quite drastic and deletes ALL stored images and containers on your entire system. In case you got a few other projects with docker images/containers that you don't want to get rid of, don't use it :).

package.json
```diff
@@ -220,28 +220,28 @@
     "build": "npm run build:web && npm run build:hyperion",
     "prebuild:api": "rimraf build-api",
     "build:api": "cross-env NODE_PATH=./ cross-env DIR=api backpack build",
-    "postbuild:api": "cp api/package.json build-api/package.json && cp now.json build-api/now.json",
+    "postbuild:api": "cp api/package.json build-api/package.json && cp yarn.lock build-api/yarn.lock && cp now.json build-api/now.json",
     "prebuild:athena": "rimraf build-athena",
     "build:athena": "cross-env NODE_PATH=./ cross-env DIR=athena backpack build",
-    "postbuild:athena": "cp athena/package.json build-athena/package.json && cp now.json build-athena/now.json",
+    "postbuild:athena": "cp athena/package.json build-athena/package.json && cp yarn.lock build-athena/yarn.lock && now.json build-athena/now.json",
```
edorivai: forgot `cp` in `... && <cp missing here> now.json build-athena/now.json`

morajabi:
@fubhy This seems to be so cool 😍

One question, what is going to happen for `src/`? Shouldn't this be a workspace subdir too?

FWIW I just adopted the same architecture by reading through this PR in a repo, it works perfectly!

morajabi:
I think we'll need to use [flow-mono-cli](https://github.com/ImmoweltGroup/flow-mono-cli) as well

api/migrations/seed/index.js
```diff
@@ -150,6 +150,8 @@ messages.map(message => {
 debug('Connecting to db...');
 // $FlowFixMe
 const db = require('rethinkdbdash')({
+  host: 'rethinkdb',
```
anarsen, line 153: This should probably read from `process.env.AWS_RETHINKDB_HOST` and then fallback to `localhost`.

seompark:
Any activity on this pull request?

morajabi:
Merging this PR doesn't make much sense given that, ZEIT Now v2 is not going to support Docker. Once Now v2 is mature enough in terms of features and tooling to migrate Spectrum to it, I suppose they'll migrate to v2, instead of Docker. 

P.S. I think not adopting Docker would help with contributions from people (from personal experience in a project)

mxstbr:
Yes, we unfortunately cannot merge this :cry:

Thanks so much for all the work here @fubhy, appreciate it anyway! :pray:

morajabi:
(@fubhy I learnt a lot from this PR <3)

jure:
> Merging this PR doesn't make much sense given that, ZEIT Now v2 is not going to support Docker. Once Now v2 is mature enough in terms of features and tooling to migrate Spectrum to it, I suppose they'll migrate to v2, instead of Docker.

Not sure I follow entirely - is this because using v2 of Now will be the recommended way to set up a dev environment? The purpose of this PR, if I understand correctly, was to speed up setting up a local dev environment. Looks pretty minimal and doesn't interact with the code - just a few Dockerfiles and one big docker-compose.yml file. 

I'm sure you have good reasons and I think you lot run the Spectrum project very well! - but from where I'm standing (someone looking at open source chat implementations), had this been merged, I'd already be playing with the platform locally. Instead, I'm having an internal debate whether it's worth setting up those services or not :)

morajabi:
@jure First, I don't run Spectrum! :) 

Just to say my opinion quick, I did implement this whole PR stack on my project which is quite like Spectrum architecture. I have these reasons to remove docker from the project but could use Docker for dbs.

- Yes, it's for dev workflow, but also was for monorepo support for deploying to Now v1. After a while now v1 will be deprecated ultimately and there will be the need for re-doing this (yet new workflow for contributors again). Also, it is for dev, but it'll be what gets deployed, it's not _only_ for dev. 
- Docker is hard to understand for a fair % people, and Docker like any tech as it's own clashing from time to time w/ issues that come up. (like some issues in this PR comments)
- It takes a huge chunk of resources (Disk, RAM, etc) while running on mac, I don't know why it's like this, and it's a known issue :( e.g. one related issue: https://github.com/docker/for-mac/issues/781

However, in my opinion might be worth considering Docker, for spinning up Redis and RethinkDB.
**Also, this is my very own experience and YMMV**, with the exact same setup as we have here. Now I run `docker-compose up` for dbs, and `yarn dev` which spins up all workers and web app with `concurrently ` NPM package.

```
"dev": "concurrently -n web,api,latte \"yarn dev:web\" \"yarn dev:api\" \"yarn dev:latte\"
```

Sorry for writing here on a closed PR. 🙌 😄

PR #2823, opened 2018-04-11
Fix token-based auth for mobile apps

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

This fixes token-based authentication, which is used by the mobile apps.
/cc @juliankrispel


mxstbr:
Admin shipping so that contributors can help us with mobile apps—otherwise they're fully blocked atm.

PR #2838, opened 2018-04-12
Fix integration tests

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

I thought I implemented that behavior, but as it turns out nodemon will actually only restart the server if a file changes.

Let's see if this fixes our integration test suite!


mxstbr:
So it seems to run fine, let's wait until our API crashes and then see if the automatic restart actually works as expected...

mxstbr:
The API still gets slow as hell the longer the tests run :confused: At the beginning the long tests take ~2s, now as we get further into the suite they take a good 15s per test.

I'm not sure whether that's an indication of our API having a problem or of our testing setup having a problem. Maybe we should try running the API in production mode somehow.

mxstbr:
Argggh, now tests started failing again because they timed out because the API is so slow—it hasn't crashed yet though. Will dig in further.

mxstbr:
The latest commit makes it so we run the production API during tests on CI. Let's hope this fixes the issue of the API taking longer and longer and eventually timing out the deeper we get into the test suite!

mxstbr:
Welp, now alllll the Cypress tests fail—that didn't work! :laughing:

mxstbr:
Oh oopsie, I'm a dummy, I didn't realize that building in a separate job than running doesn't work 🤦‍♂️ Fixed in latest commit!

mxstbr:
THE INTEGRATION TESTS PASSED!!!! :tada: The trick was to run the API in production mode, now it doesn't get slower _at all_  :100: This should mean they now pass every time, and not just every nth random time.

Todos before merging:

- [x] Update testing docs
- [x] Make API auto restart with forever
- [x] Remove old auto-restarting script and dependency

mxstbr:
This is ready—@brianlovin let's make sure to ship this asap so we can enable branch protection again and get all the existing PRs tested!

mxstbr:
Looks like there's a single test failure, and that seems to be an actual failure :tada:

brianlovin:
Hot damn dude!!! You did it! Are you looking into that last failure?

mxstbr:
I can't repro the failure locally so we're not yet 100% out of the water 🙄 I think it's just that one specific test that fails in CI though since only that one specific test failed twice in a row, so I've just commented it out for now, let's take a look at that later.

brianlovin:
Roger that.

brianlovin (approved):

brianlovin:
I'm still catching up on emails + PRs this morning, but will give it a test locally here soon, probably after we merge this.

mxstbr:
Hmm, a single test failed—again. 🤔🤔🤔🤔🤔

mxstbr:
Superseded by #2839 which adds even more awesome refactors to the CI setup.

PR #2843, opened 2018-04-12
Run integration tests against hyperion

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

**Todo**

- [ ] Make it work
- [ ] Make all tests pass
- [ ] Update testing docs


mxstbr:
@brian-mann Cypress crashes when trying to run the tests on this branch with this error:

```
{ Error: socket hang up
    at createHangUpError (_http_client.js:345:15)
    at Socket.socketCloseListener (_http_client.js:377:23)
    at emitOne (events.js:120:20)
    at Socket.emit (events.js:210:7)
    at TCP._handle.close [as _onclose] (net.js:549:12)
 code: 'ECONNRESET' }
```

...which is not very helpful :wink: I'm trying to run the integration tests against our server-side rendering server, instead of our development setup, and it all seems to be working fine except for Cypress immediately crashing when starting to run tests.

Any clues why that's happening?

brian-mann:
@mxstbr didn't see this comment until now. Is this still happening? Does it happen every time?

I see this in circle once here: https://circleci.com/gh/withspectrum/spectrum/3575?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

What else can you tell me about this? This is likely happening from the server closing the socket connection but that an error handler isn't being handled correctly in Cypress. Thus it crashes instead of either gracefully handling the error or retrying.

brian-mann:
@mxstbr if you see that it's still failing... add the `DEBUG` flag so that Cypress outputs all the logs..

`DEBUG=cypress:* cypress run`

That'll help pin down where it is failing

mxstbr:
@brian-mann no worries, thank you for taking a look at this! I added that in the last commit, you can see the full logs here: https://circleci.com/gh/withspectrum/spectrum/4450

Does anything stick out to you right before the crash?

brian-mann:
It's failing inside the browser when starting to run the tests - there is what appears to be an unhandled low level (likely net.connect) error being thrown - my guess is with the websocket connection... it appears right around the CONNECT request for `/websocket`.

Let me see if I can clone this branch and repro in either osx or docker...

Other users have reported a similar error but we've never been able to repro because it's likely a perfect storm of situations - and it could honestly be something with your server - but regardless error handling is something that is always up to Cypress to facilitate. Even in the case when its your server, we should still gracefully handle them and provide enough error context. This isn't happening in this case.

mxstbr:
Ohh I think I know what this is based on your description—our own app websocket connection fails when running the server-side rendered site locally, that could be it?!

If so, that should a) not crash Cypress and b) that error should be much more helpful :wink:

brian-mann:
I can reproduce locally. We can finally get this bug squashed. 

I believe its the same root issue as what this PR is fixing. Need to do some work on it before merging...

https://github.com/cypress-io/cypress/pull/1382

mxstbr:
I'll try and get that bug resolved so that websockets work locally when running against the SSR'd site, but I'm not yet sure whether I can. Will dig in!

mxstbr:
@brian-mann I see that PR was merged! 😻 Any chance we could get a new release of Cypress so we can take advantage of that?

mxstbr:
Picking this up now that Cypress 3.0 is out! Let's see if I can make this work :v:

mxstbr:
Oh my god after merging in #3248 this actually kinda works! Some test failures left but it's getting there!

mxstbr:
Replaced by #3334!

PR #2849, opened 2018-04-13
Fix local mobile singin

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Previously any outside contributor except for the core Spectrum employees couldn't locally log into the mobile app.

This fixes it by adding an example api token secret to the example secrets file and fixing the EXPO_URL check.

/cc @juliankrispel


mxstbr:
Shoutout to @homanp for working through this with me! 🙌

PR #2852, opened 2018-04-13
Remove Danger

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Now that we have fully moved over to Peril (hosted danger; https://github.com/withspectrum/peril-settings) we don't need to run Danger from CI anymore. (it also doesn't work for any forks)


mxstbr:
Forget it, I'll have to find a better way to do this.

PR #2868, opened 2018-04-15
Create android-friendly wrapper around draft-js-plugins-editor

mxstbr:
Instead of making the Android fix specific to the chat input this patch moves the Android fix a level deeper inside of a custom wrapper component around draft-js-plugins-editor.

This custom wrapper component conditionally renders either the standard draft-js-plugins-editor, or a vanilla text input if on Android. This should theoretically allow us to reuse the same fix for the main thread composer so folks can also create threads on Android. (that will be a follow-up PR though)

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)


brianlovin (approved):
LGTM!

PR #2883, opened 2018-04-16
Fix wrong argument for clickShareLink

Ponjimon:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- Fix a bug where it would track the wrong onboarding event


mxstbr (approved):
Oopsie, thanks!

PR #2919, opened 2018-04-19
Show photo size error on image added and neater error positioning

johnnynotsolucky:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- Previously render was only triggered once focus was lost from the input field;
- The error message when visible squashed the input field to the side
- This PR updates the ChatInput component to trigger a render when the MediaUploader onError event is hit. It also adds some styling to display the error message above the input field.

I've created a `ChatInputContainer` component to encapsulate the error and the input, and moved some of the styles out of `ChatInputWrapper`. Having two really similar names might be confusing though?

**Before**

screen shot 2018-04-19 at 21 53 01

**After** 

screen shot 2018-04-19 at 22 00 04

### DIFF
--- a/src/components/chatInput/index.js
+++ b/src/components/chatInput/index.js
@@ -21,6 +21,7 @@ import { addToastWithTimeout } from '../../actions/toasts';
 import { openModal } from '../../actions/modals';
 import {
   Form,
+  ChatInputContainer,
   ChatInputWrapper,
   SendButton,
   PhotoSizeError,
@@ -113,6 +114,7 @@ class ChatInput extends React.Component<Props, State> {
     if (currState.isSendingMediaMessage !== nextState.isSendingMediaMessage)
       return true;
     if (currState.mediaPreview !== nextState.mediaPreview) return true;
+    if (currState.photoSizeError !== nextState.photoSizeError) return true;
 
     return false;
   }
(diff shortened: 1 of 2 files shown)


brianlovin:
Wow, this is huge - thanks so much for digging in here @johnnynotsolucky; I think the changes here could definitely use some testing at mobile viewports as well, since the chat input structure has gone under a lot of iteration to make sure it behaves well on mobile. If things look okay on your end with responsive viewports, I can approve this and get it up on alpha.spectrum.chat for real-world testing :D

johnnynotsolucky:
WILCO, will feedback soon :-)

johnnynotsolucky:
There was a padding issue for mobile viewports. Thats corrected now. 

Screenies from a few of the Chrome device presets; Iphone 6, Galaxy, Ipad and Iphone 6/7/8 Plus (not in that order):

screen shot 2018-04-19 at 22 49 35
screen shot 2018-04-19 at 22 53 09
screen shot 2018-04-19 at 22 52 53
screen shot 2018-04-19 at 22 52 38

brianlovin:
This is so much better, thank you for the thorough checks!

brianlovin (approved):
Will get this up on alpha shortly!

PR #2258, opened 2018-01-10
Quoted message replies

mxstbr:
- [x] Add new field to `Message` in GraphQL called `quotedMessage`, add
resolver for it
- [ ] Build UI/UX for it

See #2142, this seems like a relatively quick feature to build that adds a ton of value for long conversations, and helps them become more manageable. I picked this up because somebody literally posted a screenshot of a message that was further up in a thread to reply to it, so it seems like this would be a useful feature. (also, apart from the Snapchat clone WhatsApp has yet to launch a single feature that didn't make 100% sense, those folks over there are :fire: with product decisions)

@brianlovin @uberbryn can one of you quickly knock out what this looks like on the front-end? Design insp with screenshots is in #2142, really doesn't need to be much.

### Will deploy
- [x] iris
- [x] hyperion
- [ ] athena
- [ ] vulcan
- [ ] mercury
- [ ] hermes
- [ ] chronos

### Run database migration?
- [ ] yes
- [x] no

## Release notes
- Quoted replies: To help you stay on top of long conversations we've added the ability to quote messages in your own message!

PR #2924, opened 2018-04-20
Message replies

### LINKED ISSUE #2142: Quoted replies
It'd be neat to be able to quote-reply to a specific message in a thread, similar to what WhatsApp does:

quoted-replies-whatsapp

That would make long threads a lot more manageable!

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)
- athena

- You can now quote previous messages in your own message to help the readability of long threads!

**TODO**

- [x] Track parent messages
  - [x] GraphQL `message.parent` field
  - [x] Store `message.parentId`
- [x] Show parent messages quoted inline
- [x] Be able to quote messages from the UI
- [x] Verify that quoted message is in same thread in the API

WIP screenshot based on designs by @uberbryn in https://www.figma.com/file/OmPWwUKus1ruaYCMrRdn2oCG/Quote%2FReply: 

screen shot 2018-04-20 at 12 02 32

This is a replacement PR for #2258 (because I fucked up a merge there and wanted to do some fundamental changes to the data model), closes #2142

Instead of going with `message.quotedMessage` (like I did in #2258) this goes with `message.parent`, which makes a lot more sense if you think about it. It'll also allow us to maybe display things in a tree structure in the future maybe potentially if we ever wanted to.


mxstbr:
@uberbryn this is basically done, I need two quick inputs from you to finish it off:

- [x] Reply icon
  screen shot 2018-04-20 at 14 02 02
- [x] How do we want to display the currently quoted message in the chat input? It's currently not displayed at all, but the information could all be there.

mxstbr:
Should quoting a message trigger a mention notification?

brianlovin:
@mxstbr yeah it probably should, but should de-duplicate if the OP is mentioned in the reply body itself.

brianlovin:
> How do we want to display the currently quoted message in the chat input? It's currently not displayed at all, but the information could all be there.

We should probably treat it similar to how we treat image uploads now; there should be an ability to cancel out of the reply.

superbryntendo:
image

preview of my changes ^

superbryntendo:
That QuoteWrapper and it's children can probably just be directly popped into the input like the pasted photos are, then. I think that'll make it pretty clear and help the experience feel consistent.

brianlovin:
Imo the left bar on the message being replied to isn’t necessary; it’s an extra visual element that adds noise when it’s already super clear that this is a reply.

superbryntendo:
@brianlovin I think it's normally clear that it's *different* from a normal message, but not immediately clear why. I think the additional clarity of purpose is worth these very minor changes.

mxstbr:
@uberbryn awesome! I made that happen in the last commit:

screen shot 2018-04-23 at 12 54 38

This is ready for a last review and then ship!

brianlovin:
@mxstbr I think we just need a way to clear it - I'm assuming that cmd+a and backspace will remove the quote? Or would it be possible to add "Cancel reply" link in there to remove it?

superbryntendo:
Getting super close, @mxstbr!

- [x] The composer doesn't clear the reply when you send the message
- [x] Offer "x" button to remove staged quotes
- [x] Remove staged quotes on backspace in empty input

We should also make sure this works ok with the photo staging state like we discussed this morning. Perhaps we should offer "x" buttons on the corners of staged quotes and staged photos? In the meantime, if you're at the beginning of the composer and hit backspace, it'd be nice as a means to clear whatever the last item staged is.

Related: what happens when you quote an emoji or image message?

superbryntendo:
Just tested quoting emoji and images. We're good 👌

mxstbr:
Staging photos and quotes together works kinda well I guess:

screen shot 2018-04-24 at 13 49 47

I'll add the same UI we have for the photo to the quote and then we should be ready to ship maybe?

mxstbr:
Latest commit makes it so that we clear both the media preview and the quoted message on backspace in the empty input, demo:



Working on the UI

mxstbr:
Last commit adds the same "X" button we have for media previews to message reply previews:

screen shot 2018-04-24 at 14 08 11

Here's what it looks like for media previews:

screen shot 2018-04-24 at 14 08 20


I think that could definitely be improved, but I'm not sure how 😅@uberbryn any ideas?

brianlovin:
Only bug I found: if I want to reply with an image, but not with any text, the flow doesn't work. Hitting the send button will send the image, but the reply will not be cleared

brianlovin:
cc @uberbryn on those styling fixes - I made it so the send button will always be at the bottom, otherwise we end up with it floating in crazy places as the chat input expands vertically.

brianlovin:
screenshot 2018-04-24 10 51 29
screenshot 2018-04-24 10 51 44

We should probably just let the media input icon stay at the bottom too, huh?

mxstbr (commented):
Looks great to me! I'll wait for @uberbryn to review the styling, then let's ship this!

mxstbr:
**TODO:**

- [x] Add e2e test(s) for this
- [x] Fix flow

mxstbr:
Found some more todos:

- [x] Inline code in quotes is invisible

  screen shot 2018-04-26 at 17 50 28

- [x] Quotes don't show up in DMs
- [x] Send mention notifications when somebody quotes a message

superbryntendo:
Good catches!
  

  
  

  
  
>

mxstbr:
Fixed those catches and deployed all changes (except for athena, i.e. you won't get mention notifications yet) to alpha!

Give it a good round of catches, and then feel free to ship it today. :100:

mxstbr:
This is working pretty great so far! I found an interesting edge case which is long messages, quoting those is really awkward:

screen shot 2018-04-27 at 13 48 07

Imagine if I sent that reply, ugh. It wouldn't look good. There's two things here that'd be great to tackle, although they shouldn't block this PR:

1. Truncate quoted messages visually, and expand them on click: (stupid mockup below, that still needs an indicator that the message was truncated and that you can click on it to expand)

  screen shot 2018-04-27 at 13 49 10

2. Be able to quote parts of messages maybe? Kinda like how you can select-and-respond on Medium we could make it so you're able to select-and-quote only part of a message. We'd probably have to make it so that clicking on it would link to the message, but imo that could be a neat feature.

Either of those shouldn't block this PR, although if you can come up with a quick fix it'd be great to get 1. done before. No biggie if not.

brianlovin:
Good catch @mxstbr - I could see two flows here exactly as you describe:

1. User highlights text of a message bubble then clicks reply - only that highlighted text is rendered in the reply
2. User replies to a long message - we truncate the plain text at ~140chars or so
3. The result of both of these is that clicking the replies *must* scroll the user to a message - the only concern I have here is that this will break in threads where we paginate upwards (e.g. load earlier messages) because it's possible that the parent message will not be loaded yet. Any ideas on how to solve that one?

brianlovin:
Hey @mxstbr and @uberbryn going to have to block this, sorry. Two more fixes needed:

1. It looks like we are missing some fields in a graphql query or something, possibly related to adding the new parent message type that doesn't have its subfields defined? 
screenshot 2018-04-27 08 28 43

2. We are not supposed to be nesting `<p>` tags
screenshot 2018-04-27 08 28 52

You can see it happening here:
screenshot 2018-04-27 08 29 02

I spent ~30mins trying to debug that, but this code is...a bit confusing on the frontend tbh.

mxstbr:
> It looks like we are missing some fields in a graphql query or something, possibly related to adding the new parent message type that doesn't have its subfields defined?

Yeah I saw that but it still works soooo... 😅 

> We are not supposed to be nesting <p> tags

Good catch, fix incoming.

----

> the only concern I have here is that this will break in threads where we paginate upwards (e.g. load earlier messages) because it's possible that the parent message will not be loaded yet.

Automatically paginate upwards until you reach the page with that message maybe?

brianlovin:
> Automatically paginate upwards until you reach the page with that message maybe?

Godspeed. Lol but for real this sounds *really* hard to pull off.

brianlovin:
One other consideration for discussion: how does this handle deleted parent messages?

mxstbr:
> how does this handle deleted parent messages?

Properly: doesn't show them. Since this isn't architected in a way where we duplicate data (COUGH NOTIFICATIONS COUGH) deleted messages are just treated as if the message didn't have a parent.

brianlovin:
> (COUGH NOTIFICATIONS COUGH)

Yeah... :(

But makes sense, thanks for testing that!

brianlovin:
Also:

> Yeah I saw that but it still works soooo... 

We should definitely fix this though, as those types of errors are the kinds of things that will break optimistic updates, and just break the apollo cache in general.

brianlovin:
Also @uberbryn this seems to be a new bug introduced with the wrong right padding:

screenshot 2018-04-27 09 01 58

mxstbr:
> We should definitely fix this though, as those types of errors are the kinds of things that will break optimistic updates, and just break the apollo cache in general.

Yeah but we don't have that data in that optimistic update and I'm not sure how we're supposed to get it :confused:

brianlovin:
I'll take another look now

brianlovin:
Yeah this is tricky; one workaround would be to get access to the store in the optimistic response object construction, but that's not possible without some hoc/closure hacking (ref https://github.com/apollographql/apollo-client/issues/1802)

I'll leave this for now I guess :/ I don't like the warnings in the console here though.

In the meantime, I found one other edge case if you try to reply to a message that was deleted (but is visible in your UI because you havne't refreshed). Working on it.

brianlovin:
Afaik this is good to go - I made a followup issue for improvements based on the above discussions #2969 

I'll get all the updates to alpha for a few more rounds of testing, can ship next week to prod!

brianlovin (approved):

PR #2935, opened 2018-04-21
Use a .prettierrc to allow editors to pick up Prettier config.

mikenikles:
This allows editors to pick up the Prettier config. Combined with a "format on save" option, this cleans up the output of `git diff` prior to committing since files get formatted with Prettier when a user saves a file.

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- Introduce `.prettierrc`. This enables code editors to pick up Prettier settings and makes the local `git diff` output cleaner before prettier formats the code as part of the pre-commit hook.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 flow-typed
+package.json
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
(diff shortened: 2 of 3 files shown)


.prettierignore
```diff
@@ -1 +1,2 @@
 flow-typed
+package.json
```
mikenikles, line 2: This is ignored to ensure npm scripts that are longer than 80 characters don't get split into 2 lines. It keeps the `package.json` file more readable if Prettier doesn't touch it.

mxstbr (approved):
Makes a lot of sense, thank you so much!

PR #2943, opened 2018-04-23
enabled content security policy

spartDev:
**Status**
- [ ] WIP
- [x] Ready for review
- [x] Needs testing

- api
- hyperion (frontend)

- Enabled CSP (Content-Security-Policy) to mitigate XSS risk by preventing resources from untrusted origins from loading.

For now I set `report-only` to `true`, to report errors in the browser but not block them. easier to test !

Later 2 options:
- Reports errors via logger but not block untrusted resources
- Block untrusted resources

I let you make me a feedback.


shared/middlewares/security.js
```diff
@@ -27,6 +28,71 @@ function securityMiddleware(server) {
...
+        // script to do data store rehydration (redux/mobx/apollo) for example.
+        // @see https://helmetjs.github.io/docs/csp/
+        (request, response) => `'nonce-${response.locals.nonce}'`,
+      ],
+
+      // Defines the origins from which images can be loaded.
+      imgSrc: ["'self'", 'www.google-analytics.com', 'spectrum.imgix.net'],
+
+      // Defines valid sources of stylesheets.
+      styleSrc: ["'self'", "'unsafe-inline'"],
```
mxstbr: `unsafe-inline` is here because we use styled-components, right?

If you set the `__webpack-nonce__` variable in `hyperion/renderer.js` styled-components should theoretically pick that up and add it to all its style tags, so we get rid of the `unsafe-inline` maybe?

spartDev: Oh that's awesome ! for sure we get rid of the `unsafe-inline` !!

spartDev: I tested but it does not seem to work

mxstbr: Hmm that's very weird. /cc @johannesnagl how did you make it work?

johannesnagl: we still have `unsafe-inline` for our intercom integration, so can't help you here unfortunately :/

mxstbr: Oh damn :confused: Thank you anyway, I guess let's just keep it with `unsafe-inline` for now.

spartDev: @mxstbr Ok Found it ! 
In fact it works great (not awake this morning 😴), set the `__webpack-nonce__` variable in `hyperion/renderer.js` apply correctly the nonce in all style tags.
But unfortunately we have also *inline styles* in the app and `styleSrc` directive includes `<style>` elements and also `style=` attributes.
We need to keep `unsafe-inline` for now. 😕

mxstbr: Ohh I see, that's fair. SGTM.

mxstbr:
I will try deploying this to alpha today and seeing how it goes!

shared/middlewares/security.js
```diff
@@ -27,6 +28,71 @@ function securityMiddleware(server) {
...
+
+      // Applies to XMLHttpRequest (AJAX), WebSocket or EventSource.
+      // If not allowed the browser emulates a 400 HTTP status code.
+      connectSrc: ['https:', 'wss:'],
+
+      // lists the URLs for workers and embedded frame contents.
+      // For example: child-src https://youtube.com would enable
+      // embedding videos from YouTube but not from other origins.
+      childSrc: ["'self'"],
```
mxstbr: We actually need this to allow every embed URL, as we allow users to embed any page they want.

mxstbr:
Deployed to `alpha.spectrum.chat` if you load that you'll see quite a few errors pop up:

screen shot 2018-04-24 at 11 44 12

I forgot that we don't upload all the users profile/cover photos to our own servers, so we also serve tons of images from Twitter/Facebook/GitHub/Google's servers.

johannesnagl:
> I forgot that we don't upload all the users profile/cover photos to our own servers, so we also serve tons of images from Twitter/Facebook/GitHub/Google's servers.

With this, it's almost impossible to come up with a whitelist for images. We experimented a lot in this area. If I would do it again, I would add a very strict CSP directly from the start of a new product, or, like in your case, if you add it later on, just come up with a very strict policy about JS/CSS and skip (almost) everything else. 

Google also notes a very script CSP here (https://csp.withgoogle.com/docs/strict-csp.html) and does not talk about images at all for example.

mxstbr:
Just leaving images open sounds reasonable to me!

spartDev:
fixed the `unsafe-eval` and open to all images.
In your screenshoot browser complain about an inline script, do you know which one is ?

mxstbr:
I don't know. Try logging into `alpha.spectrum.chat` and looking at the logs.

spartDev:
It's the google analytics tag in the header. but it's weird, I set correctly the nonce and it work fine on local 🤨

mxstbr:
Let me redeploy the latest changes, one second.

mxstbr:
Latest changes from this branch are live at `alpha.spectrum.chat`!

spartDev:
Ok it looks good except for the google analytics inline script

spartDev:
He takes the inline scripts in the `hyperion/renderer/html-templates`  right ?

mxstbr:
@spartDev maybe the problem is our ServiceWorker caching? After the first (server-side rendered) visit we serve you a static app shell HTML file (`public/index.html`) from the ServiceWorker on any request which client-side renders and doesn't ever hit hyperion. That file cannot have the correct nonce set in any of its script tags, right?

spartDev:
@mxstbr  yes that's exactly it ! When I do a curl on https://alpha.spectrum.chat/ I see the html rendered by hyperion with the right nonce.

spartDev:
For now I have no idea how to make it work, expect put the google script in a extenal file

mxstbr:
@spartDev I'd be fine with that, honestly.

spartDev:
@mxstbr done.
We need et the report-uri directive that logs the JSON detailing the violation and remove the report-only directive

oyeanuj:
@spartDev For embeds and favicons coming from `micro-open-graph`, how do you handle CSP for that? Does this line ensure that?
```
imgSrc: ['https:', 'http:'],
```

And if yes, does that also make it too broad? Or not get the most gains from a typical CSP policy?

spartDev:
@oyeanuj The `imgSrc` directive ensure that. And yes it's too wide but you can have multiple image from multiple source, all listing is impossible. Check this comment: https://github.com/withspectrum/spectrum/pull/2943#issuecomment-383874152

brianlovin:
I believe this was overtaken by #4080

PR #2947, opened 2018-04-24
#2923 Implement logic to check if new/existing is a Spectrum contributor

LetItRock:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)
- chronos

- if new user is signing in within github provider we will check if he is a Spectrum contributor and then based on that we will set isContributor flag on DBUser entity; 
- using GraphQL API its gonna be possible to check if user is a contributor (isContributor field);
(more items in future)
- cron daily job to check if some of the existing users are now contributors (there could be some users who are signed within github provider, but they are not contributing yet, so we will check every day if they are contributors now)
- implemented new OSS badge component which would be displayed in same places where is displayed PRO badge now;


api/models/user.js
```diff
@@ -174,10 +179,27 @@ const createOrFindUser = (
...
+
+      // if no user exist check if a new one is contributor
+      if (providerMethod === 'githubProviderId') {
+        return storeUserWithContributionInfo(user);
+      }
       return storeUser(user);
     });
 };
 
+const storeUserWithContributionInfo = user =>
```
littleStudent: just thinking out loud.
Maybe `storeUserWithContributionInfo` should be smt like `addContributionInfoToUser`. And look like this:
```Javascript
const addContributionInfoToUser = user =>
  isContributor(user.githubUsername, user.githubProviderId)
    .then(isContributor => {
      user.isContributor = isContributor;
      console.log(
        'Set isContributor for user: ',
        isContributor
      );
      return user;
    })
```
And the caller should handle the `storeUser(user);` and `.catch(err => storeUser(user));`?
```Javascript
if (providerMethod === 'githubProviderId') {
  return storeUserWithContributionInfo(user)
    .then(storeUser)
    .catch(err => storeUser(user);
}
```

This way we would handle the `storeUser` calls in `createOrFindUser` and not add this sideeffect to `storeUserWithContributionInfo`.

mxstbr (commented):
Apart from @littleStudent's comment above the API implementation here looks great to me so far!

I'll refer to @brianlovin for an explanation of how to add a cron job to chronos.

api/models/user.js
LetItRock: Great catch @littleStudent, thats better to move that ```storeUser``` function outside because  ```addContributorInfoToUser``` should not know about persisting and it would be more reusable... thanks!

brianlovin:
Happy to help with cron stuff, but if you're digging in on your own, I'd start with:

1. Create a new job in `chronos/jobs/index.js` - safe to run once per day, so maybe `0 0 * * *` pattern. You'll need to create a new constant string to call this job
2. Add this new job to `chronos/index` in the `createWorker` function and init the function where the other jobs are being started (46:49)
3. Create a new queue in `chronos/queues` - since GitHub has fairly strong rate limits I'd do the following:

- fetch list of all contributors to repo (1 call)
- for reach contributor, fetch that user in the db
```js
.table('users')
.getAll(GITHUB_USER_ID, {index:'githubProviderId'})
.run()
```
- do the rest of the logic to determine if any users need to be updated with a contributor badge!

Happy to help with anything here.

LetItRock:
Thank you @brianlovin for comprehensive description on the steps to implement that cron job 😃
now its more clear to me 😃I'm gonna try to implement that later today...

LetItRock:
Hello fellows 😃I've implemented that cron job based on suggestions from @brianlovin, checked that couple of times and its working as expected.
So for now only frontend part left, I need to know how we want this badge to look like, maybe @uberbryn can help me with that. And we need to decide where in app we want to show it.

LetItRock:
@mxstbr, @uberbryn, Ping... need to know how that badge should look like and where we want to show it...

superbryntendo:
Let’s make it an alternate version of the src/components/badge component that says “OSS” (not set on that but it’ll help us test). And let’s make the gradient space.alt =>  space.default and see how that all looks together.   
  

  
Let’s put it everywhere the PRO badges are to start.
  

  
  

  
  
>

brianlovin:
@LetItRock :
> Let’s put it everywhere the PRO badges are to start.

There are a few places we need to add new badges, unfortunately. Basically if you do a project search in the `src` directory you'll find everywhere we add a pro badge. But also look in `api/queries/communityMember/roles.js` - in that file we'll have to do a `loaders.user.load(userId)` to get the db record :)

Let me know if that makes sense

LetItRock:
@uberbryn, I've implemented with your suggestions and it looks pretty neat...
screen shot 2018-04-26 at 19 35 55

@brianlovin, yeah I see there are couple of places... I'll add that new badge in those... 
and later I'll try to understand what that last sentence means 😄probably it would be more clear if I check that ```roles.js``` and where it is used 😋

LetItRock:
Hey @brianlovin, one question... there is a ThreadByLine component which is displayed at the top of a thread where is basically description... and as I see from code we are only displaying here two possible types of badges "moderator" and "admin"... do we want to have there "pro" and "contributor" also?
screen shot 2018-04-26 at 22 55 32

---
Apart of that -> thats how it looks for now:
screen shot 2018-04-26 at 22 46 28

screen shot 2018-04-26 at 23 07 34

brianlovin:
@LetItRock 
> and as I see from code we are only displaying here two possible types of badges "moderator" and "admin"... do we want to have there "pro" and "contributor" also?

Yeah! Let's be as consistent as possible with badges, so want to add those to the thread byline as well?

LetItRock:
@brianlovin, cool! I'll do that later today...

LetItRock:
Thats how it looks now 😉 
screen shot 2018-04-27 at 23 25 19

api/models/user.js
```diff
@@ -166,6 +167,12 @@ const createOrFindUser = (
         }
       }
 
+      // if no user exist check if a new one is contributor
+      if (providerMethod === 'githubProviderId') {
```
alejandronanez, line 171: You can move `githubProviderId` into a constant somewhere else, that will make your code  less error prone - you could end up typing `githubProviderId` in line 171 and `gituhbProviderId` in line 186 ;-)

chronos/models/users.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+const { db } = require('./db');
+
+export const getUserWhoIsNotContributorByGithubProviderId = async (
+  githubUserId: string
+) => {
+  return await db
+    .table('users')
+    .getAll(githubUserId, { index: 'githubProviderId' })
```
alejandronanez, line 9: I guess you can reuse the constant I mentioned for `githubProviderId` in `api/models/user.js` :)

LetItRock:
@brianlovin, @mxstbr ping... I think this PR is ready to be accepted...

brianlovin:
Hey @LetItRock sorry for the slow review here - looks like there are still some `flow` failures in the branch :) This is on my list for a final review though

PR #2948, opened 2018-04-24
Update searchInput.js

dmitrika:
Removed duplicated setState with `searchQueryString: ''`.

Or it's required due to some unstated behavior?

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)


mxstbr (approved):
Hah, good catch, thank you!

PR #2951, opened 2018-04-24
Update fragments.md

alejandronanez:
You were using `userInfoFragment` instead of `storyInfoFragment` in the last `getStory` query.

**Status**
- [ ] WIP
- [X] Ready for review
- [ ] Needs testing

- Updated the api/graphql/fragments docs.


mxstbr (approved):
Thank you! We also don't have the concept of `frequencies` anymore, we should probably update that doc eventually...

PR #2959, opened 2018-04-26
chore(src): replace tab-charactor to whitespace

ryota-murakami:
**Status**
- [x] Ready for review

- hyperion (frontend)

while doing https://github.com/withspectrum/spectrum/pull/2958, I noticed that tab-charactor is not replaced whitespace by prettier.

the project's prettier settings surely use whitespace instead tab-character(prettier default), a little bit strange:thinking:


brianlovin (approved):
👌

PR #2880, opened 2018-04-16
[Mobile v1] Splash page #2329

ryota-murakami:
**Status**
- [x] Ready for review

**Deploy after merge (delete what needn't be deployed)**
- mobile

**Release notes for users (delete if codebase-only change)**
- [Mobile v1] Splash page #2329

## current progress

- [x] add github button
- [x] add button styled
- [x] fix shown bottom menu without login https://github.com/withspectrum/spectrum/pull/2980 @ArturKlajnerok
- [x] fix emoji color
- ~~[ ] fix auth login not work~~ see: https://github.com/withspectrum/spectrum/pull/2880#issuecomment-385764967 
- [x] add Code of Conduct

### screenshot
screen shot 2018-05-02 at 4 28 55

PR #2960, opened 2018-04-26
Fix/GitHub theme

ryota-murakami:
**Status**
- [x] Ready for review

- hyperion (frontend)

## Overview
i figure out this while doing https://github.com/withspectrum/spectrum/pull/2880.

in Login Page,  Github Button recived `props.theme.text.default` as a color.
i thought should use `props.theme.social.github.default` instead.
but prev `props.theme.social.github.default` assinged bluecolor('#1475DA').

so first, I confirmed that the github theme is not used elsewhere by `find src/ -type f -name "*.js" | xargs grep 'theme.social.github'`,
then i replace `theme.social.github.default` to blackcolor('#16171A')

it seems work fine on my local:computer:

## Image
screen shot 2018-04-27 at 3 00 37

thanks:smile:

### DIFF
--- a/src/components/loginButtonSet/style.js
+++ b/src/components/loginButtonSet/style.js
@@ -101,8 +101,10 @@ export const GoogleButton = styled(SigninButton)`
 `;
 
 export const GithubButton = styled(SigninButton)`
-  background: ${props => (props.preferred ? props.theme.text.default : 'none')};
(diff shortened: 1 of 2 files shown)


brianlovin (approved):
Good catch, thank you!

PR #2973, opened 2018-04-28
fix Raven.requestHandler

comus:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

from
```
requestHandler: (req, res, next) => next(),
```
to
```
requestHandler: () => (req, res, next) => next(),
```


mxstbr (approved):
Good catch, thank you!

Issue #2329, opened 2018-01-31
[Mobile v1] Splash page


mxstbr:
> Splitting up #1803 into multiple sub-issues to track against

We need a splash page you land on when you first install the Spectrum app. Some sub-todos:

- [ ] Login buttons

mxstbr:
The homepage now has the sign in links and one can sign in/out, but it's missing GitHub and it's totally unstyled.

A PR adding styling similar to the web version (https://spectrum.chat/login) would be much appreciated! :pray:

ryota-murakami:
@mxstbr sure, i'd like to take it:smiley:
i'm mid-level Web React/ReactNative dev & usually useing styled-components:nail_care:

mxstbr:
Sure @ryota-murakami, all yours! The code for this screen is in `mobile/components/Login`, and the styling should afterwards look mostly like it does on the web today:

<img width="465" alt="screen shot 2018-04-11 at 17 02 39" src="https://user-images.githubusercontent.com/7525670/38625187-2f0a4078-3daa-11e8-8c94-def619e898e9.png">

mxstbr:
@ryota-murakami let me know when you have a work in progress!

ryota-murakami:
@mxstbr i'm afraid i couldn't complete the work immediately.

current progress is.. i just added github button, but nothing any styling.(actually i'v inactive the work for a while, during weekend:sweat_smile:)
![img_0858](https://user-images.githubusercontent.com/5501268/38821565-d5fc72f4-41db-11e8-8cbe-1adce7697353.PNG)

I would like to work this week with responsibility:bow:

i just created WIP PR https://github.com/withspectrum/spectrum/pull/2880

ryota-murakami:
@mxstbr hello! this PR https://github.com/withspectrum/spectrum/pull/2880 is merge currently, but issue message's ckeckbox still empty.
is there any related task? i'd handle it if it is possile.

i'm looking forward to the mobile app:grinning:

mxstbr:
No, this is perfect! Thanks so much for tackling this!

ryota-murakami:
@mxstbr thank you:smile:

PR #2980, opened 2018-04-30
[Mobile] Signup before Tabs Screen

ArturKlajnerok:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

While looking at the app navigation setup, I've noticed that the Login screen is rendered inside Home screen. This is a simple change that will force login to be the first screen for non authenticated user.

- [x] If user is not authenticated, the login screen should be rendered.
- [x] For authenticated user let's render app tabs.

For apps that require Authentication flows on start it's a good practice to use `SwitchNavigator` as described here : https://reactnavigation.org/docs/auth-flow.html 
To use `SwitchNavigator` I would need to refactor the app initialization part a bit more, but we can start with the simple approach as proposed in this PR. 

**Before:**
screen shot 2018-04-30 at 7 32 54 pm

**After:**
screen shot 2018-04-30 at 7 32 05 pm

Login/Splash view & styling is handled here : #2880 
Parent Issue : #2329 

Let me know what do you think.
/cc @mxstbr & @brianlovin


mobile/App.js
```diff
@@ -78,13 +79,13 @@ class App extends React.Component<{}, State> {
...
 
-    const { client } = this.state;
+    const { client, token } = this.state;
 
     return (
       <Provider store={store}>
         <ApolloProvider client={client}>
           <ThemeProvider theme={theme}>
-            <TabBar />
+            {!token ? <Login /> : <TabBar />}
```
ryota-murakami, line 88: nice work @ArturKlajnerok !
I attempt to fix it in a similar way. when mentioned @mxstbr in https://github.com/withspectrum/spectrum/pull/2880

`<TabBar />` contains all bottom-ui-menu, therefore I think sounds good to switch the <TabBar /> component and <Login /> component, depending on whether or not logged in.

also i think @ArturKlajnerok mentioned SwitchNavigator is more straightforward way:smile:

ryota-murakami (approved):

mxstbr (approved):
I really like that! :100: Thanks for digging into this, let's ship it and merge it into #2880

Issue #1803, opened 2017-10-31
Mobile apps


mxstbr:
We could really use a hand at making this happen, there's quite a bit of ReactNative work to go yet. 

**If you're interested in helping out with mobile apps, please see the linked issue for the individual screens below!**

- ~~Shared component library built with `styled-components/primitives` (@brianlovin you had a WIP PR for this somewhere?)~~
- [x] Figure out if we want to use expo.io (probably?)
- [x] Set up repo architecture
- [x] Set up dev process for iOS and Android
- [x] Set up build process for iOS and Android
- [x] Decide on list of views to build for MVP
- [ ] Set up submission process to App Store and Play Store
- [ ] Build app
  - [x] Splash page, see #2329 
  - [x] Thread feeds, see #2330 
  - [x] Thread view, see #2331 
  - [ ] Notifications, see #2332 
  - [ ] Loading states (skeleton loaders like the website)
  - [ ] Null states for empty screens
  - [ ] Joining communities/channels from their profiles
  - [ ] New user onboarding (like on the website)

juliankrispel:
@mxstbr so is there a branch for this?

mxstbr:
Nope, all the WIP has been merged to `alpha`. Just continue working from there!

juliankrispel:
@mxstbr what are the services required to run the mobile app successfully?

![image](https://user-images.githubusercontent.com/1188186/38556148-f33c67e2-3cc8-11e8-8eed-be93c35713af.png)

mxstbr:
@juliankrispel hmm, should only require the API running at the same time. I think that bug exists right now, try commenting out L88-92 in `mobile/views/Notification.js`

mxstbr:
@juliankrispel that bug was fixed in #2809, and then the follow-up PR #2823 fixes authentication on mobile. Sorry for the troubles here, if you use that branch you should be good to go!

mxstbr:
I think our push notifications are working already, at least they should be—we just don't handle them yet if they come in while the app is open and don't render them properly! Ref #2332

homanp:
Ah, ok. Misunderstood then. Again I would be glad to help you out with that as well. Do you have any specific documentation on how you want to handle the notifications?

mxstbr:
@homanp moved to #2332

fullstackguyhere:
Hey @mxstbr I am interested in helping out!

PR #2989, opened 2018-05-01
[Mobile] Tab navigation styling

ArturKlajnerok:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Updated tabs styling to mach the mobile web style:
screen shot 2018-05-01 at 8 50 27 pm

- [x] Explore, Home, Message, Notification & Profile SVG Icons for tabs
- [x] Bold tab label
- [x] Enabled swipe between tabs (disabled after feedback review)

**Before:**
screen shot 2018-05-01 at 8 55 44 pm

**After :**
screen shot 2018-05-01 at 8 53 02 pm

Parent Issue: #1803


mxstbr (commented):
Yess, this is great! :tada:

I don't think we should go with swipe-between-tabs—it's only expected on Android (no app supports it on iOS), and even there it's only expected for top navs,not bottom navs. I'd vote we keep it disabled?

ArturKlajnerok:
Tabs swipe is quite useful, but it's true that it's not common on iOS. We could also enable it on Android and disable on iOS, let me know what you think about that? For now I've changed it back here  : 29073b9

mxstbr (approved):

mxstbr:
@orta huh this is interesting, it seems like `pull_request.synchronize` also runs on merge/close events, so it's not quite a replacement for Danger ootb 🤔 Maybe I need to manually filter to not include those events somehow?

orta:
I think this might be common enough to warrant a new key type

mxstbr:
Yeah I feel like if Peril is meant to replace Danger, it'd be neat to be able to do

```JS
{
  "pull_request.opened": "dangerfile.js",
  "pull_request.edited": "dangerfile.js",
  "pull_request.pushed": "dangerfile.js"
}
```

or something

orta:
Maybe 

```json
{
  "pr": "dangerfile.js",
}
```

Because that notes a strong difference between `pr` and `pull_request.*`, where pr is a special case in comparison to the others.

PR #2992, opened 2018-05-02
Archive dm threads

alejandronanez:
**Status**
- [x] WIP
- [ ] Ready for review
- [x] Needs testing

- api
- hyperion (frontend)

YES

- Initial work for `Archive Messages` mentioned in https://github.com/withspectrum/spectrum/issues/2991

So far I got:
- Mutation for marking a DM thread as `archived` / `un-archived`
- Migration to handle archiving / unarchiving


package.json
```diff
@@ -246,6 +246,7 @@
     "db:seed": "cross-env NODE_PATH=./ node api/migrations/seed/index.js",
     "db:reset": "yarn run db:drop && yarn run db:migrate && yarn run db:seed",
     "rethinkdb:migrate": "rethinkdb-migrate -f ./api/migrations/config.js",
+    "rethinkdb:migrate:generate": "rethinkdb-migrate create -f ./api/migrations/config.js --",
```
mxstbr: Nitpick: Let's call this `rethinkdb:migrate:create` in order to be consistent with the actual command run

api/migrations/20180501222559-add-isArchived-archivedBy.js
```diff
@@ -0,0 +1,31 @@
+exports.up = function(r, conn) {
+  /**
+    We're adding `isArchived` and defaulting it to `false` in `usersDirectMessageThreads`
+    All DMThreads are not archived
+  */
+  return Promise.all([
+    r
+      .table('usersDirectMessageThreads')
+      .update({
+        isArchived: false,
```
mxstbr: Do you reckon we need to do this or can we let it be `undefined` for all existing threads and handle that as if it was `false`?

api/types/DirectMessageThread.js
```diff
@@ -53,11 +53,17 @@ const DirectMessageThread = /* GraphQL */ `
...
+  input isArchivedDMThreadInput {
+    isArchived: Boolean!
+    threadId: ID!
+  }
 
 	extend type Mutation {
 		createDirectMessageThread(input: DirectMessageThreadInput!): DirectMessageThread
-		setLastSeen(id: ID!): DirectMessageThread
+    setLastSeen(id: ID!): DirectMessageThread
+    changeIsArchivedDirectMessageThread(input: isArchivedDMThreadInput!): DirectMessageThread
```
mxstbr: What do you think about calling this `archiveDirectMessageThread` or something like that in order to be more consistent with our existing mutations?

api/models/usersDirectMessageThreads.js
```diff
@@ -103,6 +104,30 @@ const updateDirectMessageThreadNotificationStatusForUser = (
...
+};
+
+const getDirectMessageThread = (
+  threadId: string,
+  userId: string
+): Promise<Object> => {
+  return db
+    .table('usersDirectMessageThreads')
+    .filter({ threadId })
+    .filter({ userId })
```
mxstbr: RethinkDB-specific performance note: Using `.filter` is reeeaaaallly slow. It iterates through the entire set before the `.filter`, so in this case this will check every single record in the table. Instead, you can use `.getAll` with an index, which is achieves the same thing much much faster!

We already have indexes for both `userId` and `threadId` on the `usersDirectMessageThreads` table, so you can use either in combination with `.getAll` and it'll be much much much faster:

```JS
return db
  .table('usersDirectMessageThreads')
  // First, use the threadId index to quickly get all records for this thread
  .getAll(threadId, { index: 'threadId' })
  // ...and only then filter by userId to get the specific record we want
  // Here, filtering is fine because here it's only 2-10 records we're filtering through, not the entire table
  .filter({ userId })
  .run();
```

We could also do it the other way around (first use `getAll` with the `userId` index, then filter by `threadId`), but my intuition is that people usually have more DM threads total than participants in DM threads, so this way _should_ be faster.

mxstbr (commented):
Looking great so far! 😍 A couple small nitpicks inline, but I'm liking the way this is shaping up

api/types/DirectMessageThread.js
alejandronanez: Makes sense to me. I will update this on my next commit.

package.json
alejandronanez: Good call.

api/migrations/20180501222559-add-isArchived-archivedBy.js
alejandronanez: Well, I chose `false` instead of `undefined` to keep things more _explicit_, but if you have strong opinions on `undefined` vs `false` I can change it.

How are handling this kind of flags across the whole application btw?

mxstbr: That makes sense to me, that's exactly how we handle them elsewhere too. Let's do this!

brianlovin:
One note, I think we might have some inconsistencies, but let's use the following convention:

```
{
  archivedAt: Date
}
```

and in graphql we can remap that to `isArchived`. 

The reason is that we want to, as best as possible, keep records of when these changes are happening, and who made the changes. This is where we aren't doing a great job, but having data models like:

```
modifiedAt: Date
modifiedBy: string
```

will give us a ton more detail than `isModified: Boolean`

brianlovin:
Also if we are unarchiving, why don't we do:

```
archivedAt: r.literal()
```

to clear the entire field? We do this for locking threads, see `api/models/thread:340-347`:

```
.update(
  {
    isLocked: value,
    lockedBy: value === true ? userId : db.literal(),
    lockedAt: value === true ? new Date() : db.literal(),
  },
  { returnChanges: true }
)
```

brianlovin:
Looking at that code snippet for locking threads, we're breaking our own rules. We shouldn't have the `isLocked` field - we should rely on the existence of `lockedAt` to determine if the thread is locked or not.

api/migrations/20180501222559-add-isArchived-archivedBy.js
brianlovin: Re: comments on the PR - why don't we just leave `isArchived` off completely? The non-existence of the field already tells us that it's not archived, and avoids us having to write a migration :)

alejandronanez: Hmmm, that said (+ your [comment](https://github.com/withspectrum/spectrum/pull/2992#issuecomment-386027520))
`archivedAt` makes sense instead of `isArchived`, but `archivedBy` does not. We already have an `userId` in `usersDirectMessageThreads` so adding `archivedBy` will be duplicating information.

brianlovin: Yup, in this case we don't need `archivedBy` :)

alejandronanez:
I'm calling the day, here's what I accomplished:

1. Created a mutation to `archive` and `unarchive` a thread
```graphql
mutation changeIsArchived($input: isArchivedDMThreadInput!) {
  archiveDirectMessageThread(input:$input) {
    id
  }
}

# variables
{
  "input": {
    "isArchived": true, # you can change true/false
    "threadId": "the thread id"
  }
}
```
2. Create the tabs for switching between `Inbox` and `Archived` messages on mobile and desktop
mobile
desktop
3. Showing a `purple dot`, this should be replace with the gear icon. Talk to @uberbryn about this. After talking with @brianlovin, we agreed that:
  a. For desktop, The dot should be visible on hover 
dot-desktop
  b. For mobile, The dot should always be visible 
dot-mobile
3. Created `/messages/archived` to load archived messages
4. Created a npm command to create migrations you should run `yarn run rethinkdb:migrate:create migration-name` in order to generate a new migration.
5. Fixed a minor bug where username was cut off in direct messages screen because to `line-height`

## What's pending:
1. Create the tooltip component and wire it up to the gear icon
2. Call the mutation using the thread id as shown in the mutation
3. Touch some code (still not sure what) in `Athena` regarding notifications, I guess this should be related to filtering what messages are not archived and send notifications only to those messages
4. Unit tests!

api/mutations/directMessageThread/archiveDirectMessageThread.js
```diff
@@ -0,0 +1,52 @@
...
+
+export default async (
+  _: any,
+  { input }: isArchivedDMThreadInput,
+  { user }: GraphQLContext
+) => {
+  const currentUser = user;
+
+  if (!currentUser) {
+    return new UserError('You must be signed in to send a direct message.');
```
LetItRock: There should be a different message right?

api/mutations/directMessageThread/archiveDirectMessageThread.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const { isArchived } = input;
+
+  try {
+    await updateIsDirectMessageThreadArchived(
+      directMessageThread.id,
+      isArchived
+    );
+
+    return directMessageThread;
```
LetItRock: I'm not sure, but probably this ```directMessageThread``` would not have updated ```isArchived``` flag...

src/views/directMessages/containers/index.js
```diff
@@ -120,6 +127,9 @@ class DirectMessages extends React.Component<Props, State> {
       data.user.directMessageThreadsConnection.pageInfo &&
       data.user.directMessageThreadsConnection.pageInfo.hasNextPage;
 
+    const isInActiveMessages = this.props.match.path === '/messages';
+    const isInArchivedMessages = this.props.match.path === '/messages/archived';
```
LetItRock: you have couple of places where you are using those strings, better to have a consts for them ;)

src/views/directMessages/containers/index.js
```diff
@@ -130,11 +140,30 @@ class DirectMessages extends React.Component<Props, State> {
...
+                to="/messages/archived"
+                onClick={() => this.setActiveThread('new')}
+                isActive={isInArchivedMessages}
+              >
+                Archived
+              </TabsLink>
+            </TabsContainer>
+            <NewMessageLink
+              to="/messages/new"
+              onClick={() => this.setActiveThread('new')}
```
LetItRock: you are creating three times arrow function which basically do the same... it would be better to declare it as a class property, and you will have only one place where the logic is...

api/mutations/directMessageThread/archiveDirectMessageThread.js
alejandronanez: Oops. Yeah you’re right!

src/views/directMessages/containers/index.js
alejandronanez: Yeah, generally speaking we have hardcoded routes all over the place. I guess it could be a good start for another PR, maybe a `good first pr` type of work.

alejandronanez: Not only routes but strings in general. Moving forward it would be great to move Errors/Messages and routes somewhere else :-)

api/mutations/directMessageThread/archiveDirectMessageThread.js
alejandronanez: Hmmm good call. We could return the updated DM from the line above to fix this. 👌🏼

src/views/directMessages/containers/index.js
alejandronanez: Yeah, I forgot to remove those before sending my last commit.

alejandronanez:
@brianlovin @mxstbr since there's no _general dropdown_ component, what do you think about show/hide an anchor that says `archive / unarchive` instead of having the gear icon that displays the dropdown on click?

alejandronanez:
I think [this](https://github.com/withspectrum/spectrum/pull/2992#issuecomment-388685525) can work for now - unless it won't delay this PR _that much_.

What I think will be a good plan for this is:
1. Finish this PR with the anchors for archive/unarchive
2. Create another PR to create a generic dropdown/pop-over component
3. Replace `archive/unarchive` anchors with the dropdown created in ^
4. Start implementing the other DM security features.

mxstbr:
That sounds good to me!

api/models/usersDirectMessageThreads.js
```diff
@@ -92,6 +92,49 @@ const updateDirectMessageThreadNotificationStatusForUser = (threadId: string, us
...
+const unarchiveDirectMessageThread = (
+  userDirectMessageThreadId: string
+): Promise<Object> => {
+  return db
+    .table('usersDirectMessageThreads')
+    .get(userDirectMessageThreadId)
+    .update(
+      {
+        archivedAt: db.literal(),
+        receiveNotifications: true,
```
alejandronanez: This should **turn notifications on** when a thread is unarchived, right? @mxstbr @brianlovin

api/models/usersDirectMessageThreads.js
```diff
@@ -92,6 +92,49 @@ const updateDirectMessageThreadNotificationStatusForUser = (threadId: string, us
...
+const archiveDirectMessageThread = (
+  userDirectMessageThreadId: string
+): Promise<Object> => {
+  return db
+    .table('usersDirectMessageThreads')
+    .get(userDirectMessageThreadId)
+    .update(
+      {
+        archivedAt: new Date(),
+        receiveNotifications: false,
```
alejandronanez: This should **turn notifications off** when a thread is unarchived, right? @mxstbr @brianlovin

api/queries/user/directMessageThreadsConnection.js
```diff
@@ -6,19 +6,22 @@ import { getDirectMessageThreadsByUser } from '../../models/directMessageThread'
 
 export default async (
   _: any,
-  { first = 15, after }: PaginationOptions,
+  { first = 15, after, isArchived = false }: PaginationOptions,
```
alejandronanez: I'm not sure about this being here (PaginationOptions)

src/views/directMessages/components/messageThreadListItem.js
```diff
@@ -1,19 +1,65 @@
...
+  };
+
+  handleArchiveDMThread = e => {
+    e.stopPropagation();
+    e.preventDefault();
+    const { thread: { id: threadId }, dispatch } = this.props;
+
+    this.props
+      .archiveDirectMessageThread(threadId)
+      .then(({ data }) => {
```
alejandronanez: @mxstbr I should update the `UserDMThreads` after this operation, but the _refresh_ is not happening. My guess is that I'm calling this mutation in `src/views/directMessages/components/messageThreadListItem.js` instead of `src/views/directMessages/containers/index.js`.
Is that correct?

alejandronanez (commented):
This is looking better 💯 
archive-dms

## TODO
- [ ] Typings
- [ ] Update DM threads after mutation
- [ ] Resolve son PR questions.

api/models/usersDirectMessageThreads.js
brianlovin: I don't think archiving and unarchiving should affect the notification settings - all it should do is hide it from the UI. New messages in that thread should definitely send notifications and mark it as un-archived right?

src/views/directMessages/components/messageThreadListItem.js
brianlovin: Not totally sure about this one - is the mutation response returning the exact same shape of data as what the direct message thread received on initial query? Apollo needs to match those in order to handle the automatic updates via cache updates.

I'm going to pull this branch and give it a try :)

api/models/usersDirectMessageThreads.js
alejandronanez: @brianlovin in the initial requirements we had:

> Archiving direct message threads: Let users archive DM threads; archived threads are hidden/not loaded for the user who archived it(!) when navigating to the main DM screen; no notifications are sent for new messages in archived threads; there's some sort of tab or filter that let's users look at their archived threads and un-archive them

As we discussed previously, we say that the only way that a message comes back to active and start sending notifications again is when you:
a) send a message in the archive thread
b) click "unarchive"

Am I missing something? I know rules can change pretty quickly though :P

brianlovin: Ah sorry, yeah I think we miscommunicated there or we had inconsistent communication - that's my bad. I think that there is a difference between muting a thread and archiving a thread. Muting would imply that you no longer want to receive notifications. Archiving is more about inbox management and clearing out old conversations to make the direct messages list easier to parse. But a new message in an archived thread is still important - so we resurface it and send a new notification.

brianlovin:
@alejandronanez I started digging in, found that pagination doesn't work. Gets this error:

screenshot 2018-05-15 14 24 37

alejandronanez:
> @alejandronanez I started digging in, found that pagination doesn't work. Gets this error:

Oh wow, I'll take a look at that then.

api/models/usersDirectMessageThreads.js
alejandronanez: If I understand this correctly is that if I want to `mute` a thread I should update  `receiveNotifications` to false and vice-versa. Correct?

alejandronanez: And muting a thread will be another task.

superbryntendo:
I'm going to work on generalizing @brianlovin's dropdown component from the settings pages so that we can use that in instances like this.

Seeing this IRL, the archive action seems a little too prominent to me - especially on mobile. Maybe we should move it inside the DM thread.

alejandronanez:
@brianlovin just in case, I fixed the pagination issue here https://github.com/withspectrum/spectrum/pull/2992/commits/9a341eb186c232c8f05d072a43d568ce4cfc99ee

mxstbr:
After yesterdays DM spam I'm picking this back up. Going to try and finish both archiving and leaving today!

mxstbr:
Latest commit finishes it off with the ability to leave DM threads. Now all three of archiving, unarchiving and leaving seem to work fine! @brianlovin mind giving this a good review both code and design wise? It's very unpolished and probably needs a little bit more work, if you just wanna jump in and finish it off 👍

brianlovin:
Phew, finally figured out how to push to this branch. :P 

Okay, so @mxstbr I didn't finish today but will be able to wrap this up tomorrow. Would appreciate some help thinking this through:

1. Archive/unarchive dm thread
2. Mute/unmute dm thread
3. Delete dm thread
4. Leave group dm thread

Here's how I envision these things behaving:

### Archiving dm threads
Archiving is used by people to triage their DM inbox. It's for people who want to clean up their inbox, see fewer old threads. This has nothing to do with spam or harassment, and doesn't impact notifications. If someone archives a DM thread, but another person replies in that same thread, it should become unarchived and send a notification.

### Muting dm threads
Muting is all about notifications. It is separate from archiving. If I mute a thread, it doesn't become archived - I simply stop receiving notifications for it. If I mute an archived thread, and someone else replies in that thread, it will unarchive the thread but I will not receive a notification for it. It will still appear as unread in my inbox. Muted threads should not badge the tab bar (this may be hard to do...)

### Leaving dm threads
You should only be able to leave a group dm thread. Eventually we'll need an "add user to thread" option for group threads, but let's not worry about it now. 

### Deleting dm threads
Deleting a dm thread deletes it for all participants of the threads. We don't delete the messages inside the thread (although we should mark them with `deletedAt: new Date()`) but simply make it impossible for any of the participants to ever see that thread in their inbox again. Any participant in a dm thread can delete that thread. This seems to be standard behavior for messaging apps.

Given all of this, I made the changes in the commits above. What I need to do next still is:
- polish up the UI on the frontend
- ensure muted threads don't send notifications
- ensure muted threads don't badge the tab bar
- test muting/unmuting
- test deleting
- test archiving/unarchiving
- text leaving a group convo

Would appreciate any help testing those cases and thinking through the backend to make sure we're making the appropriate changes in the DB to protect against querying for deleted threads, not, sending notifications for muted threads, etc.

I think a follow-on PR will be to add a 'report thread' option. For 1:1 threads it should open the report modal for the other participant. For 1:many threads it should open a modal that allows the reporting user to pick the participants they are reporting in that group thread. Let's do that later, but just trying to think through the final dropdown options:
- mute/unmute
- archive/unarchive
- leave (group)
- delete
- report

Anything else?

mxstbr:
I don't agree with all of that plan:

- **I don't think deleting should be a thing.** Why should I, a single person, be able to delete an entire group DM with 10 other people and all the content is gone not only for me but also for them? That's really really really destructive and could be used for horrible purposes like insulting somebody, waiting until they see it and then deleting the entire thread for everybody. I should be able delete DM threads for myself, but that should not affect anybody else in the DM thread—i.e. I should be able to leave it. Imo we only need leave, for all threads, no need for deleting.
- **I don't think archiving should work that way**, that's a really unnecessary feature imo. I'm never going to go and manually archive DM threads, especially if they're just going to pop back in as soon as somebody replies. (since they're ordered by recent activity anyway it's not like I'm ever digging through that list and need to organize it, where manually archiving would be useful) I could maybe see that behavior if it was automatic and would hide the thread as soon as I looked at it and either replied or ignored it, but honestly I don't want that either, I just a list of my DM threads ordered by recency.
Imo archiving should mute and hide the thread in the archived tab forever. It should never be moved back to the inbox unless I manually go into archived, look the thread up and unarchive it. That's what I'd expected when I _archive_ a thread.
- I also don't think we need to have "Report a DM thread" (yet?). In the future it might be nice to be able to report any type of content on Spectrum, but really in 99% of cases you want to report the user behind the DM thread anyway, so our report user functionality should be enough for a long while here. (especially since you can just add a comment saying you're reporting the user for that DM thread)

Does that make sense?

brianlovin:
> I also don't think we need to have "Report a DM thread" (yet?). In the future it might be nice to be able to report any type of content on Spectrum, but really in 99% of cases you want to report the user behind the DM thread anyway, so our report user functionality should be enough for a long while here. (especially since you can just add a comment saying you're reporting the user for that DM thread)

This is basically what I meant. Just an entry point from the DM thread itself to report a user :)

brianlovin:
Other concerns resolved offline:
- we won't have deleting of DMs, but will support 'leave thread' for all conversation types
- archiving will restore the thread to the inbox

brianlovin:
@mxstbr problem with "leaving": 

User A messages User B.
User B leaves the conversation.
User A messages User B again.

In this scenario, shouldn't User A see both messages, and User B only see's the second message?

This would change things quite a bit huh?

mxstbr:
Not sure, what happens when you leave a 1:1 convo in messenger and I message you again?

mxstbr:
It would just create a new conversation, right? Like ours would do too

brianlovin:
> It would just create a new conversation, right? Like ours would do too

How would this work though? User A still sees the old conversation their inbox, right? So if they reply in that thread, how would User B receive a new message without seeing the old messages? Unless that's the intended behavior?

A sends B a message.
B leaves conversation.
A sends B a message in the same thread.
B receives message and is able to see all history.

Maybe that works?

PR #2993, opened 2018-05-02
Fix CD once and for all

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Hopefully :sweat_smile:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,7 +203,8 @@ workflows:
       - deploy_alpha:
           requires:
             - test_static_js
-            - test_web
-          filters:
-            branches:
-              only: alpha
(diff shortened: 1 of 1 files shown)


mxstbr:
This works beautifully! 😍😍😍 @siddharthkp you're a genius

siddharthkp:
I'm glad it's working! ❤️

Let me know if it behaves in the coming weeks, we run into this same issue quite frequently: https://twitter.com/siddharthkp/status/980805543613816832

mxstbr:
We don't use `serve` so we'll probably not run into that :wink:

siddharthkp:
I wonder if it's a problem with `serve` though, I guess we'll find out 😄

PR #3005, opened 2018-05-03
Pageview tracking

jaydp17:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

YES

- Tracks PageViews (will come back and edit this)

**What all is done?**
- Data is stored in a tabled called `pageviews`
- Mutation to record community, channel & thread view event along with the `refererDomain`
- Query/Resolvers to get community, channel & thread event data grouped by `date` as well as `refererDomain`

**What's pending?**
- Query for popular content
- Date filter in queries
- Calling the mutation/query from the frontend


api/migrations/20180503065927-pageview-table.js
```diff
@@ -0,0 +1,13 @@
+exports.up = function(r, conn) {
+  return r
+    .tableCreate('pageview')
```
mxstbr: Let's call the table in plural `"pageviews"` like our other tables

api/types/Community.js
```diff
@@ -272,6 +272,11 @@ const Community = /* GraphQL */ `
 		id: ID!
 		customMessage: String
 	}
+	
+	input AddPageViewInput {
+		id: String!
+		referrerDomain: String
```
mxstbr: Idea: to support other types of pageviews we could add an enum `referrerType` argument here:

```JS
enum ReferrerType {
  COMMUNITY
  THREAD
  CHANNEL
}

input AddPageViewInput {
  id: String!
  referrerDomain: String
  referrerType: ReferrerType!
}
```

WDYT?

mxstbr (commented):
Looking great so far! 😍

api/types/Community.js
jaydp17: yes, it makes sense.
Then we'll move it out of community, to a separate directory, as it's no more just about community

mxstbr: Sounds good to me

api/mutations/analytics/addPageView.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+
+import { parse as parseURL } from 'url';
```
mxstbr: Minor nitpick: `url.parse` is deprecated, it's recommended to use the `URL` constructor now: https://nodejs.org/api/url.html#url_constructor_new_url_input_base

```JS
import { URL } from 'url';

// ...

const parsedUrl = new URL(url);
```

jaydp17: oh, didn't know that it was deprecated.
Thanks for that, I've updated the code.

api/models/pageviews.js
```diff
@@ -0,0 +1,67 @@
...
+};
+
+type GetPageViewParam = {
+  id: string,
+  pageviewType: PageViewType,
+};
+export const getPageViews = ({ id, pageviewType }: GetPageViewParam) => {
+  return db
+    .table('pageviews')
+    .filter({ refId: id, refType: pageviewType.toLowerCase() })
```
mxstbr: RethinkDB-specific performance note to myself: `.filter` is really slow, we should create indexes for these.

Don't worry about that though, I'll take care of it before we ship this!

jaydp17: @mxstbr does this (https://github.com/withspectrum/spectrum/pull/3005/commits/637b15a170356134127aa94e0b05f5a9ea256bb6) work?

mxstbr: Yep that totally works! 😍

PR #3038, opened 2018-05-07
fix(shared/graphql): not flow typechecked file

ryota-murakami:
**Status**
- [x] Ready for review

- shared

i figure outed not flow typechecked file.
so i enable flow typeckeck at `shared/graphql/apollo-client-options.js` & add some any type.

this file syntax seems like before es2015, so strange for me:hushed:

### DIFF
--- a/shared/graphql/apollo-client-options.js
+++ b/shared/graphql/apollo-client-options.js
@@ -1,7 +1,8 @@
+// @flow
 var apollo = require('apollo-utilities'),
   toIdValue = apollo.toIdValue;
 
-function dataIdFromObject(result) {
+function dataIdFromObject(result: any) {
   if (result.__typename) {
     // Custom Community cache key based on slug
     if (result.__typename === 'Community' && !!result.slug) {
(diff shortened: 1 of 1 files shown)


brianlovin (approved):
In this case the `any` types won't be super useful, as anything can go through here without being enforced. Might be worth typing this based on the arguments being received :)

brianlovin (changes requested):
Whoops.

ryota-murakami:
@brianlovin thanks for your review:grinning:
i'll diging this code & find out properly type.

mxstbr:
Don't worry about this too much, that file is very hard to type because essentially anything can come into it. On top of that, typing it is not very useful as you should never have to touch it—I'd probably just leave it untyped 😅 There's a lot more valuable stuff in the codebase that isn't typed!

ryota-murakami:
@mxstbr oh, i misunderstood flow convention about existing file.(reading [this](https://github.com/withspectrum/spectrum#rules))
thanks:grinning:

PR #3042, opened 2018-05-08
Encourage social media share

### LINKED ISSUE #3040: Encourage more social media shares from authors
When authors published a new thread, but haven't received a reply yet, we should show a strong upsell in the empty messages area (including buttons) to share the content to other social media sites. Dumb mockup: 

sharing upsell

> Note: we should probably come up with some better copy there, this was just a quick and dirty hack to show the idea. Don't use that exact copy!

This should only show to the author, not to anybody else—we mainly want to encourage other people to answer, rather than share.

alejandronanez:
Add method to tell wether the logged in user is the author of the thread or not

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

----

@mxstbr I added ~`lodash.get`~ `idx` so we can start doing:

~FWIW: [_.get docs](https://lodash.com/docs/4.17.5#get)~

```javascript
const foo = idx(this, _ => _.props.data.a.really.deep.nested.prop);

// instead of

const foo = this.props &&
    this.props.data &&
    this.props.data.a &&
    this.props.data.a.really ... // and so on
```

~Let me know what do you think.~

### Not author's thread
screen shot 2018-05-08 at 9 49 42 am

### Author's thread
screen shot 2018-05-08 at 9 46 29 am

Closes #3040


src/views/thread/components/messages.js
```diff
@@ -48,6 +49,7 @@ type Props = {
   data: {
     thread: {
```
mxstbr: Let's type this as `import type { ThreadInfoType } from 'shared/graphql/thread/threadInfo'; data: { thread: ThreadInfoType } }`

mxstbr (commented):
That is a very sensible suggestion! I'd prefer we use [`idx`](https://github.com/facebookincubator/idx) instead because it has 0 runtime overhead :blush: 

Other than that looks good to me so far

ryota-murakami:
I'm struggling annoying undefined check in JavaScript, it's so impressive:grinning:

alejandronanez:
> That is a very sensible suggestion! I'd prefer we use idx instead because it has 0 runtime overhead 😊

haha, np. I'll add `idx` instead of `lodash.get` 👍

src/views/thread/style.js
```diff
@@ -160,6 +163,85 @@ export const ThreadHeading = styled(H1)`
...
+export const SocialShareWrapper = styled(FlexRow)`
+  justify-content: space-between;
+  margin-top: 20px;
+
+  ${A}:first-child {
+    margin-right: 20px;
+  }
+`;
+
+export const SocialIcon = styled(Icon)``;
```
alejandronanez: @mxstbr I couldn't make this styled-component work :( that's why I had to do 

```javascript
export const BaseButton = styled.div`
 .icon {
     opacity: 1 !important;
  }

  svg {
     fill: ${({ theme }) => theme.bg.default};
  }
`
```

Instead of:

```javascript
export const SocialIcon = styled(Icon)`
  opacity: 1; // Opacity didn't change and the class for this styled-component was not applied to the Icon element.
`;
```

Maybe is something related with the `Icon` component itself?

brianlovin:
We already have a graphql type for `isAuthor` on the thread model - see `api/queries/thread/isAuthor` :)

brianlovin:
I must be missing something, because it looks like you're using that field already. Why is a new package needed here?

Sorry for my confusion!

brianlovin:
And re: styling - can we keep the same type as the empty state for messages? It should be smaller and a bit more subdued - happy to jump in and polish that up if needed.

alejandronanez:
> Why is a new package needed here?

np @brianlovin, we can use `idx` to access deeply nested properties with ease. Look this example, you can do:

```javascript
// By having this:
this.props = {
  data: {
    a: {
      really: {
        deep: {
          nested: {
            prop: 'brian'
          }
        }
      }
    }
  }
}

// You can doo:

const name = idx(this, _ => _.props.data.a.really.deep.nested.prop); // brian

// instead of

const name = this.props &&
    this.props.data &&
    this.props.data.a &&
    this.props.data.a.really ... // and so on
```

If any of those properties does not exist, you'll get `undefined` instead.

Hopefully, this makes sense.

You can read more about `idx` here: https://github.com/facebookincubator/idx

alejandronanez:
> I must be missing something, because it looks like you're using that field already.

I'm getting it from `this.props.data.thread.isAuthor`, that's correct, right?

alejandronanez:
> And re: styling - can we keep the same type as the empty state for messages? It should be smaller and a bit more subdued - happy to jump in and polish that up if needed.

Can you upload a mock please? If you're not rushed about this task I'll update the styles tomorrow or later today

brianlovin:
Ah, I see - the ability to return undefined at any point in the chain seems super useful, there are a ton of places we could use that to clean up the code!

I won't have time for a mock today, but can jump on this whenever everything else is ready to go and tidy it up :)

alejandronanez:
> Ah, I see - the ability to return undefined at any point in the chain seems super useful, there are a ton of places we could use that to clean up the code!

Yes!

> I won't have time for a mock today, but can jump on this whenever everything else is ready to go and tidy it up :)

If I finish it tomorrow (tests are still pending), do you think it will be _ok_ to merge it and iterate on it later? cc @brianlovin @mxstbr

brianlovin:
The UI will just need some polish before we merge it in - should just take a second when this is ready :)

alejandronanez:
@brianlovin Ok, I completed my work on this, when you have the mocks ready I can come back and finish this task :)

I'll see what I can grab next in the meantime 👍

src/views/thread/style.js
superbryntendo: We shouldn't use the Icon component here - we should use the Button component from src/components/buttons and set the glyph prop to the icon you prefer.

alejandronanez: Thanks for the heads up @uberbryn. 👍

alejandronanez:
I just got rid of `FacebookButton` and `TwitterButton` components from `style.js` after knowing the existense of `<Button gradientTheme={'social.facebook'}` in [this pr](https://github.com/withspectrum/spectrum/pull/3082/files#diff-aefb8bed155fa774112ad531678c665aR82). Thanks @uberbryn !

mxstbr:
Sorry for the super slow work here, I'm picking this back up! Will merge conflicts, maybe adjust the styling a little bit and then shipit 🚢

mxstbr:
Adjusted the styling and copy a little bit:

screen shot 2018-08-06 at 14 00 48

Now it fits well with the other upsell, looking similar:

screen shot 2018-08-06 at 13 58 27

Pending a quick review by @brianlovin this is ready to go!

alejandronanez:
Yay!

I’m glad to see this moving forward! :-)

brianlovin:
@alejandronanez this is super close! Last thing is it just needs a tiny bit of padding on mobile, or perhaps the buttons should stack vertically on mobile so that the bbuttons don't run up against the left and right edges of the screen. Then let's ship and iterate!

mxstbr (approved):
Donezo, took care of that really quickly! Shipping :ship:

brianlovin:
@mxstbr dumb workflow question, but when you commit directly there are you doing it on the fork? Or do you have a faster way to get involved in a contributor branch?

mxstbr:
Doing it on the fork! Using https://github.com/github/hub makes it a hell of a lot easier:

```sh
# Checkout the PR #3042 (this one)
hub pr checkout 3042
# ...work on stuff...
git commit -m 'Bla'
# Push to the fork
git push git@github.com:alejandronanez/spectrum.git alejandronanez-encourage-social-media-share-issue\#3040:encourage-social-media-share-issue#3040
```

That last `git push` is the hardest part: you have to push to their fork, but locally the branch is called `<username>-<branchname>`, where on the fork it's obviously only `branchname`, so we use the `:` delimiter to push our locally named branch onto a different branch in the fork. Not sure if this makes sense, but it's the fastest way I've found of doing this!

PR #3065, opened 2018-05-11
Create "Off Topic" channel for new communities

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

This should help create a better sense of community. /cc @uberbryn
@brianlovin

### DIFF
--- a/api/models/channel.js
+++ b/api/models/channel.js
@@ -252,6 +252,22 @@ const createGeneralChannel = (
   );
 };
 
+const createOffTopicChannel = (communityId: string, userId: string) => {
+  return createChannel(
+    {
+      input: {
+        name: 'Off Topic',
+        slug: 'offtopic',
+        description: 'Random banter and off-topic discussions',
+        communityId,
+        isPrivate: false,
+        isDefault: false,
+      },
+    },
+    userId
+  );
+};
+
 const editChannel = async ({
   input: { name, slug, description, isPrivate, channelId },
 }: EditChannelInput): Promise<DBChannel> => {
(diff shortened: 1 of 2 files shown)


mxstbr:
"Random" might be a better name potentially maybe? Any opinions?

superbryntendo:
I read the paper, so I know why you think this is a good idea, but I'm just not sure I agree. 

Some thoughts:
1. I'm a little skeptical of how much the research applies because the community used as an example in the paper wasn't freeform the way Spectrum is. It was basically just movie ratings, so naturally it didn't form community without somewhere for members to communicate.
2. Structuring casual, off-topic conversation into threads seems somewhat high-friction for achieving this goal, but this seems like a pretty common use of the watercoolers from what I'm seeing. Maybe the ideal solution is just to optimize those?
3. If we wanted to go the channel route, what do you think about just offering this as a kind of "channel template" in onboarding with some additional context as to why it's valuable? I'm a little uncomfortable with automatically generating them bc I suspect most communities probably wouldn't want to have that without some idea as to why it's beneficial.

mxstbr:
> this seems like a pretty common use of the watercoolers from what I'm seeing. Maybe the ideal solution is just to optimize those?

The problem with watercoolers is that newcomers to the community jump in and ask their questions there. Two reasons, it's way less friction to post a message than post a thread and they are very obviously exposed to everybody. That means it stops being a shoot-the-shit-banter-kinda area for oldtimers immediately.

That's why I went with the non-default channel route, that way it's very unobvious and only folks who've been in the community for a while will know it exists. No newcomers! I agree that maybe a whole channel is a bit much, but it's a quick and necessary improvement imo.

> I'm a little uncomfortable with automatically generating them bc I suspect most communities probably wouldn't want to have that without some idea as to why it's beneficial.

As a couter datapoint to that, Slack auto-creates `#random` without any explanation or opt-out functionality—you gotta delete it if you don't want it! I don't feel like that's going to be confusing honestly, but I'm open to ideas

PR #3066, opened 2018-05-11
Reset thread.lastActive if the last messages is deleted

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

This fixes a bug where thread would show up at the top of the feed even though it seems like no new messages were sent. Users could send messages, which would catapult the thread to the top of the feed, and then immediately delete them so nobody would see them but the thread would still stick to the top of the feed.


brianlovin (dismissed):

brianlovin (dismissed):
Nice! This looks good

PR #3069, opened 2018-05-11
Add hover profile to mentions

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- Hover profiles on mentions!

**EDIT**: This works now!!!!


src/components/rich-text-editor/style.js
```diff
@@ -4,6 +4,65 @@ import styled, { css } from 'styled-components';
...
+      <div onMouseEnter={this.hover(true)} onMouseLeave={this.hover(false)}>
+        <Manager>
+          <Reference>
+            {({ ref }) => <Link to={`/users/${username}`}>{children}</Link>}
+          </Reference>
+          {this.state.hovered && (
+            <Popper
+              placement="top-start"
+              modifiers={{
+                preventOverflow: { enabled: false },
```
FezVrasta: you probably want to keep this on?

mxstbr: Still happens with `preventOverflow` set to `enabled: true` :confused: (I copy and pasted this from the other place we use react-popper)

src/components/rich-text-editor/style.js
```diff
@@ -4,6 +4,65 @@ import styled, { css } from 'styled-components';
 import Link from 'src/components/link';
 import { Transition, zIndex } from '../globals';
 import { theme } from 'src/components/theme';
+import { Manager, Reference, Popper } from 'react-popper';
+import HoverProfile from 'src/components/avatar/hoverProfile';
+import { getUserByUsername } from 'shared/graphql/queries/user/getUser';
+
+const MentionHoverProfile = getUserByUsername(
+  props =>
+    console.log(props) ||
```
alejandronanez: Oops, you forgot to remove this one :)

src/components/rich-text-editor/style.js
FezVrasta: I'll be able to play with it during the weekend if it's okay to you.

On top of my head I may suggest to define the `boundariesElement` to make sure it's set to the right container, but it should be done automatically by popper.

brianlovin (approved):
Noice!

PR #3091, opened 2018-05-14
Push notification settings

isaiahgrey93:
PR adds granular controls for web push notificaitons alongside current email settings.
https://github.com/withspectrum/spectrum/issues/2831

**Status**
- [x] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

I am unsure still to what all will need be deployed.

Re: WIP status - I would like feedback on whether a batch request would be preferable here. I am currently toggling off each value for web push that is enabled when web push is globally disabled.

The result is that you see momentarily each checkbox for web push become unchecked until disappearing while globally disabling web push notificaitons.

**BEFORE**
before

**AFTER**
after

**AFTER_MOBILE**
after-mobile


mxstbr:
Picking this up since we'll want it for the mobile apps...

**TODO**

- [ ] Store and show push notification settings even if web push is disabled
- [ ] Filter mobile and web push notifications in athena based on users push notification settings
- [ ] Filter desktop push notifications based on users push notification settings (ref #2825)
- [ ] TBD

PR #3118, opened 2018-05-19
Update CSS transitions

alejandronanez:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

I changed almost all of css transitions that were like `transition: all ...` for `transition: specific-property ...`. This will tell the browser "I want you to listen to only background-color and opacity, and animate those properties" instead of "Hey browser, listen for every single property that could change and then animate it"


mxstbr (approved):
This is a very very very good idea 👌 Thank you!

PR #3124, opened 2018-05-20
[Flow] express dispatch is redux dispatch(action)​ method

ryota-murakami:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

## overvirew

update all `dispatch: Function` type definition to `dispatch: Dispatch<Object>`.
`Dispatch` type came from `import type { Dispatch } from 'redux';`.

thanks🙏


brianlovin (approved):
Nicely done, thanks for continuing to improve the coverage here!

PR #3168, opened 2018-05-29
Flaky cypress test

tadasant:
Tiny PR: was looking through the E2E tests and this one kept failing locally. If it wasn't failing the CI build previously, I imagine it's due to a difference in machine specs.

Problem was that the `cy.get('[data-cy="join-community-button"]').should('be.disabled');` was getting invoked too late. Sequence that occurred:

1. Test clicks button, invokes async request
2. Button becomes disabled
3. Async request returns, button becomes enabled
4. Test begins waiting for button to become disabled

Sometimes happened on the first `.should('be.disabled')`, sometimes on the second. With the changes, haven't experienced any failures.

Screenshot of failure prior to change:

flaky test

**Status**
- [ ] WIP
- [X] Ready for review
- [ ] Needs testing

N/A

### DIFF
--- a/cypress/integration/community/view/profile_spec.js
+++ b/cypress/integration/community/view/profile_spec.js
@@ -167,7 +167,6 @@ describe('public community signed in without permission', () => {
       .contains(`Join ${publicCommunity.name}`)
       .click();
 
-    cy.get('[data-cy="join-community-button"]').should('be.disabled');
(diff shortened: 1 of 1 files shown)


mxstbr (approved):
Ohh thank you so much for digging into that! ❤️

PR #3204, opened 2018-05-30
🚧 Implement reply to selected text feature

dev-drprasad:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

ref #2969

### DIFF
--- a/src/components/chatInput/index.js
+++ b/src/components/chatInput/index.js
@@ -389,6 +389,7 @@ class ChatInput extends React.Component<Props, State> {
         messageType: !isAndroid() ? 'draftjs' : 'text',
         threadType,
         parentId: quotedMessage.messageId,
+        snippet: quotedMessage.snippet,
         content: {
           body: !isAndroid()
             ? JSON.stringify(toJSON(state))
@@ -543,6 +544,7 @@ class ChatInput extends React.Component<Props, State> {
           messageType: 'media',
           threadType,
           parentId: quotedMessage.messageId,
+          snippet: quotedMessage.snippet,
           content: {
             body: reader.result,
           },
--- a/src/components/chatInput/index.js
+++ b/src/components/chatInput/index.js
@@ -42,7 +42,12 @@ const QuotedMessage = connect()(
   getMessageById(props => {
     console.log('props :', props);
     if (props.data && props.data.message) {
-      return <QuotedMessageComponent message={props.data.message} />;
+      return (
+        <QuotedMessageComponent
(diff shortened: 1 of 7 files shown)


src/components/message/index.js
```diff
@@ -73,6 +73,7 @@ class Message extends Component<Props> {
       replyToMessage({
         threadId,
         messageId: message.id,
+        snippet: window.getSelection().toString() || null,
```
mxstbr, line 81: Ohh TIL about `window.getSelection`, that's very neat! My main question about this approach is: how can we make sure that whatever is quoted is actually in the message that's being quoted?

dev-drprasad: @mxstbr I commented about it in issue https://github.com/withspectrum/spectrum/issues/2969#issuecomment-393448330

src/components/chatInput/index.js
```diff
@@ -40,8 +40,14 @@ import type { Dispatch } from 'redux';
 
 const QuotedMessage = connect()(
   getMessageById(props => {
+    console.log('props :', props);
     if (props.data && props.data.message) {
-      return <QuotedMessageComponent message={props.data.message} />;
+      return (
+        <QuotedMessageComponent
```
dev-drprasad, line 46: @brianlovin @mxstbr i am modifying `QuotedMessage` to selected text. is it good idea ?? should we create seperate component like `QuotedSnippet` ??

src/components/chatInput/index.js
```diff
@@ -89,7 +95,7 @@ type Props = {
   networkOnline: boolean,
   threadData?: Object,
   refetchThread?: Function,
-  quotedMessage: ?{ messageId: string, threadId: string },
+  quotedMessage: { messageId?: string, snippet?: string },
```
dev-drprasad, line 102: @brianlovin @mxstbr  are we using any library for types ?

src/components/message/view.js
```diff
@@ -160,7 +170,7 @@ const Action = (props: ActionProps) => {
             tipText={`Reply`}
             tipLocation={'top'}
             size={24}
-            onClick={replyToMessage}
+            onMouseDown={replyToMessage} // onClick deselects selected text
```
dev-drprasad, line 173: @mxstbr @brianlovin `onClick` deselects selected text, so we always get `null`. so, i had to use `onMouseDown`. what do you think ? is there any better way ?

PR #3222, opened 2018-05-31
Fix desktop auto updating maybe

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Instead of relying on a bunch of custom behavior this patch makes it so we use the standard `autoUpdater.checkForUpdatesAndNotify()` method and call it every 15 minutes.

It also adds a menu item called "Check for updates" so users can manually check for updates.


brianlovin (approved):
Awesome! Let's give it a try

PR #3287, opened 2018-06-11
Mobile misc cleanup

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- mobile

----

- Cleans up top bar titles across all mobile views
- Moves time formatting util used in mobile from `src` to `shared`


brianlovin (approved):
This looks good to me - I think 'Loading {entity}...' feels a bit awkward in practice, and maybe it can just be left blank or we put some kind of loading indicator there? 

Not high pri, but I'll look into it as we polish. Regardless, everything else looks good here :)

PR #3298, opened 2018-06-12
Fix viewing new profiles on mobile 

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Viewing a profile that didn't have any data loaded for it yet was broken due to `navigation.state.params` being undefined. This patch switches us to using the standard `.getParam` method which accounts for this.


brianlovin (approved):
TYVM!

PR #3314, opened 2018-06-13
Mobile buttons

### LINKED ISSUE #3313: [Mobile] Proper buttons
Right now we just use the standard react-native `<Button />` across the board, but that doesn't look very great honestly:

screen shot 2018-06-13 at 11 06 34

We should re-create our button setup from the web client.

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review 
- [ ] Needs testing

- mobile

**Related issues (delete if you don't know any existing issue(s) that could be related to this PR)**

Closes #3313

### DIFF
--- a/mobile/components/Button/index.js
+++ b/mobile/components/Button/index.js
@@ -1,0 +1,82 @@
+// @flow
+import React from 'react';
+import TouchableHighlight from '../TouchableHighlight';
+import ConditionalWrap from '../ConditionalWrap';
+import Loading from '../Loading';
+import type { GlyphTypes } from '../icon/types';
+import type { Node } from 'react';
+
+import styled, { css, withTheme } from 'styled-components/native';
+
+const ButtonView = styled.View`
+  justify-content: center;
+  align-items: center;
+  border-radius: 8px;
+  padding: 12px 16px;
+  background-color: ${props =>
+    props.color ? props.color(props) : props.theme.brand.alt};
+
+  // States
+  ${props =>
+    props.state === 'disabled' &&
+    css`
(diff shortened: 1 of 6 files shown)


mxstbr:
This is done—I'm happy with the component API!

It still doesn't _look_ perfect by any means (especially with icons) and doesn't cover everything we'll need, buuuut it's enough to use it throughout the app and then iterate on as we go along and need different states.

This is the API, critique very welcome!

```
type Props = {
  onPress: () => any,
  title: string,
  state?: 'disabled' | 'loading',
  size?: 'large',
  color?: (props: { ...Props, theme: Object }) => string,
  icon?: GlyphTypes,
  // NOTE: Passed in via withTheme, no need to pass manually
  theme: Object,
};
```

brianlovin:
@mxstbr it's really nice! I made a couple small tweaks on the styling + layout, and renamed the `title` prop to `label` as that seems to be a bit more ergonomic when composing button props. But agreed, open to iteration here for sure and adding more types as needed! Thanks for doing this, looks great so far :)

brianlovin (approved):

mxstbr:
> and renamed the title prop to label as that seems to be a bit more ergonomic when composing button props

Fwiw the reason I went with `title` in the first place is because the core ReactNative `<Button />` uses title—I don't really care though :blush:

brianlovin:
Oh damn, didn’t see that. Okay I’ll switch that back for better dx, thanks for the context!!

mxstbr:
No worries, really not that big of a deal.

PR #3299, opened 2018-06-12
Mobile Thread Composer

mxstbr:
**Status**
- [ ] WIP
- [ ] Ready for review
- [ ] Needs testing

**Deploy after merge (delete what needn't be deployed)**
- mobile

**Related issues (delete if you don't know any existing issue(s) that could be related to this PR)**

Closes #3293

**Todo**

- [x] Open thread composer as modal
- [x] Add basic wireframe of elements in the thread composer
- [x] Create `<Select />` component for community and channel picker
- [x] Fetch user communities and channels and render pickers
- [x] Add text input for title and body
- [x] Publish on button press
- [ ] Proper styling

<details>
  <summary>Click to see gif of current status</summary>

</details>

PR #3316, opened 2018-06-13
Mobile DM composer

### LINKED ISSUE #3255: [Mobile] Compose direct message

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- mobile

**TODO**

- [x] Add modal navigator with DM composer screen
- [x] Add chat input
- [x] Let users search and select users to send the DM to
- [x] Create DM thread and then redirect to it

Note @brianlovin: I'll try and reuse your `PeopleSearchView` and `SearchInput` components, which will likely require a bit of reworking to be more reusable. Will see how I get along with that!

Closes #3255


mxstbr:
This is done! :tada: Uses the same modal navigator thing as the thread composer (ref #3299) and reuses the existing `PeopleSearchView`, also already has the new null state built-in.

Here's a demo:



I realize it's not perfect styling-wise, but it should be good enough for now.

mobile/views/Search/PeopleSearchView.js
```diff
@@ -19,8 +19,9 @@ type Props = {
     search: SearchUsersType,
   },
   ...$Exact<ViewNetworkHandlerProps>,
-  navigation: NavigationProps,
+  onPress: (userId: string) => any,
```
mxstbr, line 21: @brianlovin this is the most interesting change code-wise. Instead of automatically redirecting the to the user profile when clicking on a user search result I added an onPress prop, which means I can fully reuse this existing component in the dm composer! 😍😍😍 (I made it go to the user profile on the search view, obviously)

mxstbr:
Added an X symbol to the selected user pills so it's clear that clicking them will remove the user:

screen shot 2018-06-13 at 16 55 09

brianlovin:
Are you still working on this or is it ready for review? It has both the WIP and ready for review labels :P

mxstbr:
It's ready for review! Sorry about that, seems to be a bug with Peril

brianlovin:
@mxstbr I pushed up a bunch of changes, going to merge this since it's much nicer and works well.

Notice https://github.com/withspectrum/spectrum/pull/3316/commits/7079da1866130980cd66723485d16198b8903971 => this will allow people to init a DM from anywhere in the app. The problem I ran into tonight, and spent ~1hr working through but couldn't solve a bunch of edge cases for, was that we should really have both the DMComposer and ThreadComposer be globally accessible modals in the navigation stack.

Imagine: 
- init a DM composer from a user profile, from a long press on a thread, from a thread detail view
- init a Thread composer from the home feed, community feed, channel feed

I tried moving these modals up to a global level in our navigation but ran into edge cases like:
- If you init a DM from the home feed (by long-pressing a thread item), it should push to the DM thread view upon submit, but the back button from the DM thread view should return the user back to the home screen
- If you init a DM from in the DM tab, it should push to the DM thread view upon submit, but the back button from the DM thread should return back to the DM tab.

I couldn't get this behavior to work, and knew it would only get more complex when thinking about initing DMs from user profiles and other places. So for now that commit above has the behavior:
- If you init a DM thread, then submit, it will push to the DM thread view, and the back button will return you to the DM tab.

Hope that makes sense. Anyways, might be something good to pair on or sync on tomorrow. Either way, this is coming along nicely!

brianlovin (approved):

PR #3334, opened 2018-06-15
Run integration tests against hyperion

mxstbr:
**Status**
- [ ] WIP
- [ ] Ready for review
- [x] Needs testing

Replaces #2843, based on #3243, also based on #3382 (so until that's merged you'll also see those changes here)

This basically means we're running integration tests in the production setup, which is a big deal. :tm: We should (in a follow-up PR) also be able to configure Cypress to fail tests on SSR checksum mismatches (in case that doesn't happen yet) so we catch those as soon as they happen instead of months after the fact by chance!

- [x] Make Cypress run against hyperion
  - [x] in CI
- [ ] Make all the tests pass
  - [x] 3 failing in `community_settings_billing_spec.js`
  - [x] 1 failing in `home_spec.js`
  - [x] 1 failing in `inbox_spec.js`
  - [ ] 2 failing in `navbar_spec.js`
  - [x] 1 failing in `thread_spec.js`
- [ ] Update docs


mxstbr:
They are runninnnng!!!

mxstbr:
@brian-mann CircleCI errors out because "Too long with no output (exceeded 10m0s)", but there was Cypress output to the console the whole time 🤔 Any clues why that happens?

brian-mann:
It means something happened when trying to spawn the browser to run `thread/delete_spec.js...`

Doesn't make sense considering all of the other specs spawned correctly. Rerun CI a few times to see if it continually happens.

You can also try setting the DEBUG logs which will output a massive stream of data but will give details as to where Cypress was when this happened.

`DEBUG=cypress:* cypress run`

Another option is to try Chrome although that won't record a video. We automatically detect things like Electron crashes, and we automatically handle when the browser **doesn't** connect (we will correctly time out). So what I'm thinking here is that the browser actually **did** launch, but that something failed to load related to Cypress, since it doesn't appear that your tests ever began.

Try rerunning so we can see if it's just that spec file.

mxstbr:
Now they ran through, must've been a fluke! Thanks for the response 🆗

brian-mann:
Other users are reporting the same problem such as this: https://github.com/cypress-io/cypress/issues/1934

There is something we need to look at and fix.

mxstbr:
All of the remaining tests fail due to the same error: "SecurityError: Blocked a frame with origin 'http://localhost:3006' from accessing a cross-origin frame." Unfortunately, the stack trace is totally useless in debugging why this happens:

screen shot 2018-06-18 at 13 52 16

Interestingly, this happens to all the tests that load the inbox view. @brianlovin do you have a hunch what this could be? What is the inbox view loading that could cause this, are we redirecting somewhere?

@brian-mann have you seen this before/could that error be made to be a bit more helpful? Right now, I'm stuck here, not knowing why this is happening or what I have to do to fix it.

mxstbr:
I thought disabling chrome web security would fix it, but apparently it didn't... 🤔 

EDIT: Nevermind, now only the navbar specs are failing!

mxstbr:
I think the remaining tests are failing because the ServiceWorker is caching HTML pages instead of going to the server-side rendering server or serving the app shell. 

@NekR we want to do the standard app shell pattern where the first visit to a page returns the server-side rendered HTML, then the SW caches an app shell (`/index.html`) and then all future visits return that app shell, which loads fresh data from the API. How can we tell the offline-plugin to not cache the server-side rendered HTML, and only the app shell? What are we doing wrong?

https://github.com/withspectrum/spectrum/blob/760da30868e8038f60e232473cea771e1dc8d6fb/config-overrides.js#L106-L120

mxstbr:
@nekr that last commit fixes it: a4f345d 

The issue was that our app shell lives at `/index.html`, which the default `rewrites` method turns into `/`—but the app shell doesn't get returned for `/`, only for `index.html`! I think the `appShell` option shouldn't be passed through rewrites, that was very very annoying to debug? Or at least document the default `rewrites` method and make it clear that `appShell` is being run through it.

mxstbr:
Now I'm getting "Timed out waiting for the browser to connect." for all tests 🤔

mxstbr:
Since it seems like it'll take a bit longer to figure out Cypress 3.0 I've moved the ServiceWorker fix to #3384, which we should ship asap.

NekR:
Yeah, that makes sense. I'll do the explanation and add warning that
appShell is passed through rewrites.

"Not passing it through rewrites" is a question for a major release though.
If you're interested in that, please file an issue in the repo 🙏

P.S. Glad you were able fix the issue and sorry for the poor docs.

PR #3350, opened 2018-06-19
Make mobile app build on Android

sjchmiela:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Closes https://github.com/expo/expo/issues/1753

**Note**
When starting the React Native packager one has to clear the cache for this change to be reflected in the bundle—here we're using Expo, so `exp start -c .` (the `-c` flag has to be included just once).

### DIFF
--- a/mobile/.babelrc
+++ b/mobile/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["babel-preset-expo"]
+}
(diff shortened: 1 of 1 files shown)


mxstbr (commented):
This might need a couple of the plugins and presets we use in the root babelrc too? (specifically thinking of the GraphQL stuff)

sjchmiela:
Oh yes, sorry, I didn’t investigate the codebase further. 😕 I could try and figure out which plugins from the root are needed in mobile too if you’d like (but I guess it would take you significantly less time than me. 😃)

mxstbr (approved):
Thank you so much for figuring this out, much much appreciated! :pray:

PR #3351, opened 2018-06-19
Mobile app Android fixes

### LINKED ISSUE #3352: [Mobile] Make all screens run on Android

### LINKED ISSUE #3353: [Mobile] Make it look good on Android

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- mobile

Man, ReactNative on Android is _fun_. :roll_eyes:

**Crashing**

- [x] Fix DraftJS rendering
- [x] Fix thread view

**Styling**

- [x] Fix Avatar backgrounds
- [x] Fix header buttons
- [x] Fix stacked avatars
- [x] Fix message bubbles and text
- [x] Fix reactions
- [x] Fix ChatInput
- [ ] Fix search view

Closes #3352, closes #3353


mxstbr:
Last commit(s) add the info icon to mobile and change the header action buttons to be icons. Looks like this on Android:





And like this on iOS:

screen shot 2018-06-20 at 11 27 16

As I said in the commit message, I'm happy to add text-based top right buttons back on iOS in case they feel more native or something. (I use Android so no clue)

mxstbr:
`overflow: visible;` doesn't work on Android.

>  overflow: visible only works on iOS. On Android, all views will clip their children.

Ref https://facebook.github.io/react-native/docs/layout-props.html#overflow, plus these issues https://github.com/facebook/react-native/issues/6802 and https://github.com/facebook/react-native/issues/17074 

This means we have to rework the way stacked avatars work, digging into this.

mxstbr:
Stacked Avatar should be possible with absolute positioning, much worse is the DraftJS rendering being broken due to nested `Text`s. (which don't work on Android, but do on iOS) Thinking through how we're going to fix that, might require an upstream patch to `redraft` where we can say that `Text`s should never be nested? /cc @lokiuz

mxstbr:
DraftJS turned out to be an easy fix, I'm a dummy! Just had to add the `String.fromCodePoint` polyfill 👌

mxstbr:
@brianlovin bit my teeth out on stacked avatars, so I'm just going to un-stack them on Android for now. Search view is totally borked on Android, but we're going to change that anyway, so I'll also leave that alone for now:

mxstbr:
Last thing that's really fucked is Messages:

screenshot_20180620-151252

Two main issues, first the same as with the stacked avatars that everything on Android is `overflow: hidden`, and second the new Text component is a bit borked there I think?

mxstbr:
Fixed messages and a bunch of other small things! Now the only thing that's still borked is the search view, but I'm going to tackle that another time once we know what we're doing with it.

@brianlovin this is ready for a review & merge!

brianlovin (approved):
Thanks so much for digging through all these!

PR #3380, opened 2018-06-22
Fix SSR bugs

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- Fixed a bunch of misc bugs

**Todo**

- [x] Chat input sometimes merging with "join" upsell
- [x] DM threads list no scrollable
- [ ] ~~Thread subscribed status not updating in real-time~~

### DIFF
--- a/src/views/directMessages/components/threadsList.js
+++ b/src/views/directMessages/components/threadsList.js
@@ -80,7 +80,6 @@ class ThreadsList extends React.Component<Props, State> {
           hasMore={hasNextPage}
           loader={<LoadingDM />}
           useWindow={false}
-          initialLoad={false}
           scrollElement={scrollElement}
           threshold={30}
           className={'scroller-for-community-dm-threads-list'}
--- a/src/views/thread/container.js
+++ b/src/views/thread/container.js
@@ -232,20 +232,6 @@ class ThreadContainer extends React.Component<Props, State> {
 
     if (isBlockedInChannelOrCommunity) return null;
 
-    const LS_KEY = 'last-chat-input-content';
-    const LS_KEY_EXPIRE = 'last-chat-input-content-expire';
-    let storedContent;
-    // We persist the body and title to localStorage
-    // so in case the app crashes users don't loose content
-    if (localStorage) {
-      try {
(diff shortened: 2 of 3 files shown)


mxstbr:
The thread subscribed status not updating in real-time has nothing to do with SSR. It always works in the inbox view, and it never works on the thread view, no matter whether it's client-side rendered or server-side rendered. That makes sense if you think about it, because what updates the thread data once you're subscribed? The `listenToNewThreads` GraphQL subscription that runs in the inbox view—which does not run on the thread view.

We can fix that by running a subscription on the current threads' channel when looking at the thread view, but let's postpone that fix because it's not a big deal. That leaves the DM threads list not infinite scrolling, will dig into that!

src/views/directMessages/components/threadsList.js
```diff
@@ -80,7 +80,6 @@ class ThreadsList extends React.Component<Props, State> {
           hasMore={hasNextPage}
           loader={<LoadingDM />}
           useWindow={false}
-          initialLoad={false}
```
mxstbr, line 83: @brianlovin this prop `initialLoad={false}` is used throughout the app
in more places—and I assume it breaks infinite scrolling after SSR in
all of those places? Why was that introduced in the first place, what's
it needed for? Removing it fixes the issue and doesn't _seem_ to break
anything, but I'm not sure if I'm just missing something...

brianlovin: Honestly no clue...let's try removing it and test on alpha for a few days.

brianlovin:
Last thing here would be figuring out why we're seeing 9 day old notifications in the dropdown:

screenshot 2018-06-22 10 57 25

And when we see these stale notifications, the 'mark all as read' button doesn't work

mxstbr:
> Last thing here would be figuring out why we're seeing 9 day old notifications in the dropdown:

Do you happen to have a consistent repro for that by chance? :angel:

Let's just ship this for now to solve these annoying bugs and then tackle the notification one in a separate PR? (I don't even know if that's a SSR thing, and it's not worth blocking this PR for that fix)

brianlovin (approved):

PR #3382, opened 2018-06-23
Fix dashboard SSR

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

Ref #3303

@brianlovin 63ad628db _should_ fix the `?t` query param even when SSR'ing I think? I haven't been able to verify because I can't seem to log into the SSR'd site locally anymore. (not sure why) Further down below we check `(hasThreadsButNoneSelected || loadedNewThreads)` before selecting the first thread, which means even if one was selected as soon as new ones load in it would select the first one, right? So this new condition should make sure that doesn't happen if we mounted with that thread.

### DIFF
--- a/shared/middlewares/thread-param.js
+++ b/shared/middlewares/thread-param.js
@@ -2,7 +2,7 @@
 
 const threadParamRedirect = (req, res, next) => {
   // Redirect /?t=asdf123 if the user isn't logged in
-  if (req.query.t) {
+  if (req.query.t && !req.user) {
     res.redirect(`/thread/${req.query.t}`);
     // Redirect /anything?thread=asdf123
   } else if (req.query.thread) {
--- a/src/index.js
+++ b/src/index.js
@@ -22,45 +22,32 @@ import { track, events } from './helpers/analytics';
 import { wsLink } from 'shared/graphql';
 import { subscribeToDesktopPush } from './subscribe-to-desktop-push';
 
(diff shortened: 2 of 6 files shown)


mxstbr:
Deploying to alpha for testing!

brianlovin:
Hm, looks like this isn't working on alpha, right?

mxstbr:
It does fix one thing: selected threads are now correctly SSR'd! It still selects the first feed thread after loading the feed, but the selected threads now SSR as they should.

Will dig into why it's still selecting the first feed thread.

mxstbr:
Look at these logs:

screen shot 2018-06-25 at 11 14 30

We unset the mounted thread ID before any threads were loaded, so when they finally do load in `mountedThreadId` is null and the app thinks it has the select the first thread! Will figure out how to fix this.

mxstbr:
The last commit fixes it, the dashboard now works beautifully as far as I can tell! 😍 Deploying to alpha for testing again.

mxstbr:
Working on one last problem: the loading state in the thread feed after SSR is finished.



Note how it shows the loading state even though all the threads are already loaded—it should just be doing that in the background.

mxstbr:
Fixed the loading state too!!! 😍 The dashboard is very very nice now:



No loading state no nothin', it all just works as it should!

brianlovin (approved):

PR #3463, opened 2018-07-02
Update to draft-js-markdown-plugin v3

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Fixes all the pasting issues by removing the custom paste behavior, also fixes inline styles mixed with inline code thanks to @andrewfan


brianlovin (approved):
🙌

PR #3494, opened 2018-07-05
Messages redesign

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Opening up a PR for this so we can discuss these changes. (this is currently live on alpha.spectrum.chat in case you want to see what it looks like)


mxstbr (commented):
Nitpicks:

- [ ] Badges look a bit off, at least on Firefox
- [ ] Inline code needs a slight border
- [ ] It's not very clear whether you've liked a message before or it just has one like from somebody else

brianlovin:
Fixed those nits. Gathering more feedback.

brianlovin:
Made some more changes, getting latest fixes on alpha

cypress/integration/thread/chat_input_spec.js
```diff
@@ -103,7 +103,7 @@ describe('chat input', () => {
       cy.visit(`/thread/${publicThread.id}`);
     });
 
-    it('should allow quoting a message', () => {
+    it.only('should allow quoting a message', () => {
```
mxstbr: that's one way to make e2e tests pass 😜

brianlovin:
Whoops!

cypress/integration/thread/chat_input_spec.js
mxstbr: nit: gotta make sure all e2e tests run

shared/graphql/queries/thread/getThreadMessageConnection.js
```diff
@@ -76,6 +77,19 @@ export const getThreadMessageConnectionOptions = {
...
+    // the messageId from the hash and start pagination from there. This
+    // allows users to share links to individual messages and the pagination
+    // will work regardless of if it's a super long thread or not
+    if (
+      props.location &&
+      props.location.hash &&
+      props.location.hash.length > 0
+    ) {
+      variables.after = props.location.hash.substr(1);
+      variables.last = null;
```
mxstbr: Should probably also set `variables.first = 50`, right? Just to make sure.

mxstbr (commented):
Code-wise this looks pretty good to me apart from two small nitpicks. The code is much neater than it was before too, so good job on that!

----

Design-wise, after using it exclusively for a bit I think we should go with this over the bubbles! :+1:

Both reading and taking action on messages is so much nicer than with the bubbles, and while it was a bit jarring for me initially I got used to it. In fact, I noticed I hate using prod because it's so uncomfortable to read there, especially for long threads with lots of rich formatting. (also the timestamps-per-message are so helpful)

Design nitpicks:

- [ ] The new mention styling is too much in-your-face imo, it draws too much attention to itself and detracts from reading the rest of the message
- [ ] The timestamp next to the author byline on the first message is kinda in the middle of nowhere, feels out of place / like it has some wrong margin
- [ ] Message timestamps should link to the message link? Just like threads do
- [ ] Author bylines should show the hover profile too?

mxstbr:
Oh and we gotta fix the loading indicators, they don't look so great:

screen shot 2018-07-07 at 18 29 26

On top of that, you first get the shimmer when the whole thread view is loading, then you get the thread title and the spinner for the body and the messages, and then you get the messages but still the spinner for the body and then you get the body. It's... not great.

brianlovin:
Got it, working on this today. I'm just refactoring the hover profile logic to be applicable to any element easily. Got a pretty solid fix, but is requiring me to fix a lot of unflowed files :/

mxstbr:
Another small design nitpick:

- [ ] Messages are less apart then paragraphs within a single message, e.g.

screen shot 2018-07-08 at 22 43 15

Note where the first message ends and how there's more space between the individual paragraphs of a single message than between the messages:

screen shot 2018-07-08 at 22 43 09

Should probably reduce the margin between paragraphs?

mxstbr:
Maybe this makes it clearer?

screen shot 2018-07-08 at 22 43 15

The two boxes are two messages. Note how the margin between each paragraph within a message is much bigger than the margin between the messages.

brianlovin:
I gotcha! Will dig in this week and finish cleaning all this up. My current weekend work was fixing and abstracting out hover profiles

brianlovin:
> The new mention styling is too much in-your-face imo, it draws too much attention to itself and detracts from reading the rest of the message

I think I agree with everything except this - I appreciate seeing usernames highlighted in a different way so I can see who is talking back and forth way more clearly. For people who don't quote messages but just @ reply it makes it easier to follow the context if the usernames can be easily scanned. Happy to play with styling overall, but would not want to revert to it being just a pure link :)

mxstbr:
I totally agree, (especially the "current user was mentioned yellow" is super useful), I think it just needs to be toned down a notch for all other mentions—right now it detracts from reading a bit.

brianlovin:
@mxstbr I *think* I've taken care of everything you pointed out, and then some. Feel free to peruse my latest commits there from the last couple of days. I'm putting all of this on alpha now.

Would love feedback specifically on the hover profile abstraction.

cc @uberbryn I refactored out the avatar components to be either a `UserAvatar` or a `CommunityAvatar` which dropped the api surface area down considerably. Open for feedback here as well :)

brianlovin:
One thing I need to test on alpha is updating the user profile/cover photo. I'm having trouble with it locally but not on prod (regardless of branch), so there might just be some funky behavior with image uploading locally in general.

brianlovin:
Image uploading for profile/cover photos seems to be working on alpha; must be a busted local thing then :)

Would love more feedback here and we can leave it on alpha for a couple days team!

brianlovin:
Found one bug on mobile where the `hoverProfile` container components are preventing things from being tapped. Will fix tomorrow.

brianlovin:
- [x] allow hover profile components to be clicked-through on mobile
- [x] prevent clicking on reactions ui in messages from propagating events to select message
- [x] double check that the proper messages are being selected when loading a thread url with a hash message id
- [x] allow click outside of message to unselect
- [x] remove hover state emphasis on selected message

brianlovin:
Oh boy, turns out node doesn't have access to hash fragments. So my method of selecting messages didn't work with SSR. I've migrated to using an `&m=:id` query parameter for selected messages. As far as my local testing goes, this works with SSR and client-side navigation. Pushing to alpha.

brianlovin:
cc @mxstbr for review on https://github.com/withspectrum/spectrum/pull/3494/commits/f01ecdbf64aa713de9b46145284d4524ef28acfc

brianlovin:
That test failure is v weird: 

screenshot 2018-07-13 14 20 24

brianlovin:
Oookay so the dependency bump is actually failing to compile. Will revert versions and just flow ignore the problematic line.

brianlovin:
Duh: https://github.com/sindresorhus/query-string/releases/tag/v6.0.0

Reverting to lower version

brianlovin:
Added lots of integration tests in https://github.com/withspectrum/spectrum/pull/3494/commits/6967c74fa085971057df66af4f1d8d4942ff674b - let me know if you can think of anything else we should cover here @mxstbr :)

brianlovin:
https://github.com/withspectrum/spectrum/pull/3494/commits/416aacee58ca5384d111cc747676586907d62ff1 fixes so many issues with the avatar component by not maintaining our own code and relying on `react-image` which has not only fallback states, but also handles entire image load failures (e.g. bad filetype, 404s, etc)

src/components/avatar/image.js
```diff
@@ -1,203 +1,48 @@
...
+          {...this.props}
+          decode={false}
+          loader={
+            <Fallback
+              size={size}
+              mobilesize={mobilesize}
+              type={type}
+              src={fallbackSrc}
+              alt=""
+            />
```
mxstbr, line 33: Why don't we just show nothing while loading the avatars? I think that'd make it feel more like the native image loading of the browser, it feels kinda weird to show the default while loading?

mxstbr (commented):
Maaaan this is soooo great 😍😍😍😍 Love it!

src/components/avatar/image.js
brianlovin: Cleaned this up to add a subtle placeholder, agree on flashing the default being weird

brianlovin:
https://github.com/withspectrum/spectrum/pull/3494/commits/b46a14235a134d982bff427c1d3bc94f78d7b426 makes a change to the way messages are selected that was frustrating to me: basically I said any click on the message should select it. This was weird because selecting it didn't really...mean anything. And it broke highlight text on an unselected message. Instead, I've made it so that timestamps and the 'copy link' action will select the message and act as a url to that specific message. I think this is more clear, but putting it on alpha for a bit more testing.

I think this will allow me to remove most of the `e.stopPropagation` calls I was forced to make, which will simplify this code considerably.

mxstbr:
That makes sense to me.

brianlovin:
Ah crap, I remember why I had that behavior now: mobile. Needs a way to expose the actions ui on mobile. Will think on this today

brianlovin:
Kk, should be fixed! Will have this ready to go tomorrow.

morajabi:
Hooray!

brianlovin:
Woot woot! We're live :)

PR #3495, opened 2018-07-05
Community Analytics v1.1

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

The goal of this patch is to make community analytics more findable and to show what it can do to make it more enticing to pay for it!

**TODO**

- [x] Make analytics viewable by all team members
- [x] Always show analytics item in dashboard sidebar
- [x] Show growth stats without paying
- [x] Show upsell within all other cards (top members, top convos, new convos)
- [x] Adjust upsell styling to fit within cards
- [x] Fix individual upsell copy
- [ ] Fix e2e tests?
- [ ]  TBD


mxstbr:
Current status:

screen shot 2018-07-05 at 11 10 53

Next step is to make the upsell fit within the cards, now that it's no longer full screen.

mxstbr:
I think it'd be cool if the background of the upsells looked "blurry/see-through" and showed a fake preview of what the card will look like once unlocked, if that makes sense? Not sure if that's too much work though, probably not worth it...

mxstbr:
Did a quick MVP refactor of the analytics upsell, but it's obviously not shippable in this state and I'm unsure where to take it from here:

screen shot 2018-07-05 at 12 10 35

@brianlovin would you mind taking a look at this?

brianlovin:
@mxstbr spent some time thinking on this, and can't get past the fact that having 3 upsells on screen for the same thing just feels super weird and confusing. People might wonder if each one does something different, or if each of those modules costs $100/mo each.

I don't mean to undo a lot of your work here, but here's a solution I ended up on:
screenshot 2018-07-05 22 12 07

This makes the upsell fit in with exposing the growth cards for free users. This might be a good middle ground? Again, not sure what's best here, except that 3 upsells for the same thing probably isn't the best design choice :)

LMWYT or any changes you want!

mxstbr:
I just feel like we need to give people a taste of what they:re going to get for their money, which the three cards did

brianlovin:
> I just feel like we need to give people a taste of what they:re going to get for their money, which the three cards did

I see where you were going with it, for sure. But three empty cards there only to show the titles feels weird. One thing we did at Buffer back in the day was mock up a "blurred out" version like you'd mentioned - but I think that will have less impact with our current analytics design since we don't have any charts or anything like that...

Not sure tbh.

mxstbr:
Maybe the copy just needs to be more clear and have bullet points?

Upgrade now and get insights on:

- **Weekly top members**
- **Weekly top conversations**
- **Unanswered conversations**

Or something?

brianlovin:
screenshot 2018-07-06 12 16 51

Updates, wdyt

mxstbr (commented):
LGTM, let's shipit

mxstbr:
Apparently we never shipped this, let's fix that!

PR #3700, opened 2018-08-03
Fix CORS configuration

### LINKED ISSUE #3698: Cross Origin Resource Sharing(CORS) misconfiguration leads to Same Origin Policy(SOP) bypass
**Severity** : Medium

**Description** : Usually one website cannot access the contents of another, because the browser has something called as Same Origin Policy. But there are other ways to access the data using CORS headers like Allow-Control-Allow-Origin and Allow-Control-Allow-Credentials. In this vulnerability, when the Origin was sent as a header to the spectrum API while making an AJAX request, it successfully blocks the access by removing the Allow-Control-Allow-Origin header, however it was checking for pre-domain wildcards, which lets an attacker to access the data using a special domain like attacker-spectrum.chat. [More in Explanation section down below]

**Proof of Concept** : [Video Link](https://mega.nz/#!YrxixAxQ!UKSTzedhoQI2V90XOQuLnCXSgQIp44wGihewbq_dYLk)

**Reproduction steps** : 

1. Download the html file(cors-exploit.html) attached to this email.
2. Add **127.0.0.1 attacker-spectrum.chat** to your hosts file. For windows its under C:\Windows\System32\drivers\etc\hosts file and for linux its /etc/hosts. The reason for this is that we don't have to buy a domain "attacker-spectrum.chat" and this is a required step because spectrum.chat server checks for *spectrum.chat as its origin instead of *.spectrum.chat, it thinks its checking for all subdomains, but it missed a dot, which lets anyone to register a domain with suffix of spectrum.chat and make this exploit work.
3. Setup a simple server, it could be apache, xampp or anything as simple as `python3 -m http.server 80` or even just open the html file directly in a browser.
4. Make sure you are logged into spectrum.chat, you don't need spectrum.chat opened in another tab or anything, but make sure you are logged into it and the browser holds the cookie.
5. Visit `attacker-spectrum.chat` in the browser and wait for 2-5 seconds, you'll see that your sensitive details from user.json is being dumped onto the webpage, which clearly illustrates sensitive data being exfiltrated by another website without authorization.

**Explanation** : 

- As you know, under account settings, a user can download his profile related information (endpoint: `https://spectrum.chat/api/user.json`)
- If we visit the endpoint and look at the response, we can see that there are CORS headers like Allow-Control-Allow-Origin and Allow-Control-Allow-Credentials.
- Since it's a AJAX request, the browser automatically sends the origin header to the endpoint and in the response we see the origin being reflected under Allow-Control-Allow-Origin header.
- But If we try to change the origin to something else, the server completely removes the Allow-Control-Allow-Origin header.
- For example: `Origin: https://spectrum.chat` will give you the response of `Allow-Control-Allow-Origin: https://spectrum.chat`, but if the origin header was `Origin: https://attacker.com`, the server responds with no _Allow-Control-Allow-Origin_ header at all.
- If the Origin was something like `Origin: https://arbitraty-sudomain.spectrum.chat` the server responds back with the CORS header like `Allow-Control-Allow-Origin: https:// arbitraty-sudomain.spectrum.chat`, so there's a check on the server-side.
- The server checks if the origin is a valid main domain(spectrum.chat) or a subdomain(*.spectrum.chat). But here comes the problem, the server doesn't properly check for subdomain in the origin. Instead of check for `*.spectrum.chat`, it checks for `*spectrum.chat`.
- As you can see there's a dot missing, which can let an attacker to register a domain like `attacker-spectrum.chat` and make this exploit work, because `attacker-spectrum.chat` is valid with the rule `*spectrum.chat`.
- In the video and reproducing steps, I've used etc hosts file so that I don't have to buy attacker-spectrum.chat domain(although it's [available](https://www.namecheap.com/domains/registration/results.aspx?domain=attacker-spectrum.chat))

**Impact** : An attacker can simply register a domain `*spectrum.chat`(ex: _hacker-spectrum.chat, legit-spectrum.chat, whateverspectrum.chat_) and when a authenticated user visits the malicious webpage, the attacker can steal sensitive information without authorization and the user wouldn't even know that this happened.

**Mitigation** : Setup proper rules while checking for subdomains, check for `*.spectrum.chat` instead of `*spectrum.chat`.

**References** : 
- [Wiki](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing)
- [PortSwigger Blog](https://portswigger.net/blog/exploiting-cors-misconfigurations-for-bitcoins-and-bounties)
- [Detectify Blog](https://blog.detectify.com/2018/04/26/cors-misconfigurations-explained/)
- [GeekBoy Blog](https://www.geekboy.ninja/blog/exploiting-misconfigured-cors-cross-origin-resource-sharing/)

Have a great day!
s0cket7

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

Closes #3698

@s0cket7 does this look correct to you?

### DIFF
--- a/shared/middlewares/cors.js
+++ b/shared/middlewares/cors.js
@@ -4,11 +4,7 @@ import cors from 'cors';
 export default cors({
   origin:
     process.env.NODE_ENV === 'production' && !process.env.FORCE_DEV
-      ? [
-          'https://spectrum.chat',
-          /spectrum-(\w|-)+\.now\.sh/g,
-          /spectrum\.chat$/,
-        ]
+      ? ['https://spectrum.chat', /\.spectrum\.chat$/]
       : [/localhost/],
   credentials: true,
 });
--- a/shared/middlewares/cors.js
+++ b/shared/middlewares/cors.js
@@ -4,7 +4,11 @@ import cors from 'cors';
 export default cors({
   origin:
     process.env.NODE_ENV === 'production' && !process.env.FORCE_DEV
-      ? ['https://spectrum.chat', /\.spectrum\.chat$/]
+      ? [
+          'https://spectrum.chat',
+          /\.spectrum\.chat$/,
+          process.env.NOW_URL,
+        ].filter(Boolean)
       : [/localhost/],
   credentials: true,
 });


PwnFunction (approved):
Yeah this looks fine.

brianlovin (approved):

PR #3701, opened 2018-08-03
Enable host validation in API, fix CSRF

### LINKED ISSUE #3688: Cross-site request forgery(CSRF) vulnerability
**Severity** : High

**Description** : Since there is no valid way of determining where the request to spectrum is originating from, an attacker can setup a malicious website and when the victim visits that page, the attacker can make any number of authenticated requests to spectrum as the victim.

**Proof of Concept** : [Bug Demo Video](https://mega.nz/#!o6YTmSrL!sKCL0KNFbQPKeehsLKyI1p6CBdlJere8dQHH5PWtibM)

**Reproduction Steps** :

1. Log into your spectrum.chat account and check your account setting like name and website.
2. Visit `https://devangeles.com/csrf-exploit.html` and wait for 2 seconds.
3. Go back to your settings and refresh(CTRL+F5 - no cache refresh), you'll see that your name, image, and other things have changed to something different(Nothing harmful for the sake of proof of concept, only basic settings change). If you didn't see the changes try clearing the cache and then reload.
4. If you want to try it out locally, I've attached the exploit file(_csrf-exploit.txt_) down below, rename it to _whatever.html_ and open in a browser to see the exploit work.
[csrf-exploit.txt](https://github.com/withspectrum/spectrum/files/2240131/csrf-exploit.txt)

**Explanation** : 

- Whenever there's a request made to the sever which updates something or does something which should only be done by an authenticated user, it's always necessary to have valid way of detecting the origin of the request or else a malicious website can send the same request and server thinks it came from your website, but it didn't.
- Since a website can make requests to any other website, but cannot read data due to Same Origin Policy, CSRF attacks don't break this rules at all, since they only send the request and don't really care about the response.
- Browser happily sends the cookie to the website whenever a request is made regardless of the origin. So there has to be a way to validate that the request came from the expected origin and we do that using random tokens. We generate a strong random token and send it as a part of the html form or something like that. When the request is made, we get back the same token and we validate if they are the same. If it was not the same token, then we drop the request.
- In the case of spectrum, there was no CSRF tokens, which lets an attacker to simply repeat the request (captured by a web-proxy like burp suite) from a malicious domain as an AJAX request and make things happen.

**Impact** : An attacker can make any request to an endpoint in the context of a victim's session like delete the any account without authorization.

**Mitigation** : Find a good way to identify and validate the origin request, probably using randomized tokens. [More on Mitigations](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet)

**Extras**

- JSON CSRF : Since the endpoint needed a _multipart-form content-type_, it was easy to forge the request, but when it comes to a _JSON content-type_, it's a little different. We need to prepare somethings for this exploit to work. This basically works on spoofing the content-type via a **Flash file(_.swf_)** and then make a **307** redirect to the endpoint to make the CSRF work. If you want to try deleting the user account, this method is perfect because the content-type has to be _application/JSON_ and there's no _multipart-form_. I've also made a proof of concept for the JSON CSRF attack which you can find it here `https://devangeles.com/JSONCT/`. Open the `exploit.html` and allow flash to work. This exploit does not delete the user's account, I thought that would be really dangerous to try out, so instead, this proof of concept is for changing the name and website of you account settings. See references down below to know more about JSON CSRF. 
[JSON CSRF Demo Video](https://mega.nz/#!py5CRIKK!SeVF02TGaykcYTH1G3fFjk5n7CGltwzrxIn_8WYAhBA)
- Recently I reported a CORS misconfiguration bug to @brianlovin over an email, got some great feedback, Thanks for that, but anyways, If we pair these 2 exploits, which is CSRF + CORS misconfiguration, it's an Arbitrary READ (CORS) + Arbitrary WRITE (CSRF) of data to a victim's account and that's really dangerous. I hope the team fixes these issues quickly.

**References**

- [Wiki](https://en.wikipedia.org/wiki/Cross-site_request_forgery)
- [OWASP](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF))
- [JSON CSRF](https://www.geekboy.ninja/blog/tag/json-csrf/)

Have a great day!
s0cket7

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

In combination with #3697, this should fix #3688. (if this doesn't break
everything)

This definitely needs testing on alpha, it might break all the things.

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -12,6 +12,7 @@ import Raven from 'shared/raven';
 import { ApolloEngine } from 'apollo-engine';
 import toobusy from 'shared/middlewares/toobusy';
 import addSecurityMiddleware from 'shared/middlewares/security';
+import csrf from 'shared/middlewares/csrf';
 import { init as initPassport } from './authentication.js';
 import type { DBUser } from 'shared/types';
 
@@ -31,6 +32,8 @@ app.use(toobusy);
 
 // Security middleware.
 addSecurityMiddleware(app);
+
+app.use(csrf);
 
 // Send all responses as gzip
 app.use(compression());
--- a/api/index.js
+++ b/api/index.js
@@ -33,7 +33,9 @@ app.use(toobusy);
 // Security middleware.
 addSecurityMiddleware(app);
 
-app.use(csrf);
+if (process.env.NODE_ENV === 'production') {
+  app.use(csrf);
+}
 
 // Send all responses as gzip
 app.use(compression());
--- a/api/index.js
+++ b/api/index.js
@@ -33,7 +33,7 @@ app.use(toobusy);
 // Security middleware.
 addSecurityMiddleware(app);
 
-if (process.env.NODE_ENV === 'production') {
+if (process.env.NODE_ENV === 'production' && !process.env.FORCE_DEV) {
   app.use(csrf);
 }
 
--- a/api/package.json
+++ b/api/package.json
@@ -53,6 +53,7 @@
     "highlight.js": "^9.10.0",
     "history": "^4.6.1",
     "hoist-non-react-statics": "^2.5.5",
+    "host-validation": "^1.1.0",
     "hpp": "^0.2.2",
     "imgix-core-js": "^1.0.6",
     "immutability-helper": "^2.7.1",
--- a/api/yarn.lock
+++ b/api/yarn.lock
@@ -3792,6 +3792,10 @@ home-or-tmp@^2.0.0:
     os-homedir "^1.0.0"
     os-tmpdir "^1.0.1"
 
+host-validation@^1.1.0:
+  version "1.1.0"
+  resolved "https://registry.yarnpkg.com/host-validation/-/host-validation-1.1.0.tgz#59e8ee8cd3573f644279ed25475ad9865b7042d4"
+
 hosted-git-info@^2.1.4:
   version "2.5.0"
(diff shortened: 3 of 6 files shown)


mxstbr:
Deployed as https://build-api-bpujbgcvcs.now.sh, let's see if this works there, then I'll alias this to alpha!

mxstbr:
Well shiat:

screen shot 2018-08-03 at 15 33 08

Digging in

mxstbr:
Lots of experimentation and deploys later, I finally figured out the right combination! :tada: Here's a deploy, this should work on the deploy preview and then I'll alias it to alpha and that should also work!

~~https://build-api-qvdwlsagaz.now.sh~~

mxstbr:
Latest commit allows all subdomains, something I missed. Here's a new deploy with those changes: https://build-api-eswfiwycmh.now.sh

This deploy preview should redirect you to prod, and if it works I'll alias it to `api.alpha.spectrum.chat` which hopefully doesn't break alpha. 🤞

mxstbr:
Works on the deploy preview, aliasing to alpha.

mxstbr:
Aliased to `api.alpha.spectrum.chat` and alpha still works! :tada: This is good go I think! 👍

brianlovin (approved):
Good work! Thank you for digging in on this :)

PR #3734, opened 2018-08-08
Add zeit.co to CORS whitelist

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

Necessary to query the public API from zeit.co. /cc @rauchg @leo


brianlovin (approved):
OH DANG!

PR #3778, opened 2018-08-16
Enable Apollo Engine again and cache Community.metaData

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

This sets Apollo Engine up again and makes it cache public responses in the server memory.This is obviously not optimal as we want the cached result to be returned by all servers, not every server having to cache! Not sure yet what to do about that because they don't support Redis, only memcached...

It also sets up the cache directive for Community.metaData, which should theoretically mean we only recompute that every 30 mins or so.

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -94,25 +94,24 @@ import createSubscriptionsServer from './routes/create-subscription-server';
 const subscriptionsServer = createSubscriptionsServer(server, '/websocket');
 
 // Start API wrapped in Apollo Engine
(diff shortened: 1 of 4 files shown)


brianlovin (approved):
LGTM - seems worth it to have it on alpha for a couple hours today before hitting prod :)

mxstbr:
Yep sounds good!

PR #3796, opened 2018-08-18
remove Flow version specify

ryota-murakami:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Closes https://github.com/withspectrum/spectrum/pull/3795

I saw this PR https://github.com/withspectrum/spectrum/pull/3795 failing by `.flowconfig` value.
I remove that value section In order to accept above PR.

Flow is not involved actual code, so use least version is fine!
I have a fascination to Flow I'd like to maintain it.(I'm reading OCaml book and attempt hacking flow😎)

Thanks😀

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -39,6 +39,3 @@ untyped-type-import=error
 untyped-import=warn
 unclear-type=warn
 unsafe-getters-setters=error
-
-[version]
-0.66.0


mxstbr (commented):
The reason we have this in the config is so that people with v78 on their local machine don't run into all sorts of errors when running on our codebase. We do need this there I'm pretty sure!

ryota-murakami:
@mxstbr Oh I see, that is so noisy.
I'm waiting for https://github.com/facebook/flow/pull/4916 I want to only focus crucial part of all!

Thanks😀

PR #3798, opened 2018-08-18
Frecency thread feeds (second try)

### LINKED ISSUE #3718: Top Threads

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- mercury
- api
- hyperion (frontend)

Deploy in that order!!

Supersedes and thus closes #3727
Supersedes and thus closes #3726
Closes #3718

Based on the conversations in #3727 here is a new approach to frecency thread feeds.

**TODO**

- [x] Add new `calculateThreadScore` queue
- [x] Process jobs in mercury, assigning a `score` and `scoreUpdatedAt` for each thread
- [x] Add jobs to the queue from mutations changing properties of a thread related to the score
- [ ] Figure out what to do with all the existing threads
- [x] Add `sort` argument to `Community.threadConnection` GraphQL type
- ~~Add `sort` argument to `Channel.threadConnection` GraphQL type~~
- [x] Look at results with recent db backup and tweak algorithm if necessary
- [ ] After merging first deploy mercury to production and then the API and the frontend to alpha, then check alpha to see if all looks fine


mxstbr:
These are the results on a not-very-recent prod backup.

**Before:**

screen shot 2018-08-18 at 19 21 57

**After:**

screen shot 2018-08-18 at 19 23 52

I don't really like these results much yet. The current algo indexes heavily on recent (i.e. last activity), but that means that a thread with 150 participants that's half a year old will shoot up to the top if a single person writes another message. :confused:

I think we need to do time boxing when considering the raw content, and count messages that were sent a while ago a lot less than messages that were sent just now. We can do that now! Will dig into that.

mxstbr:
I edited the algorithm to exponentially bias towards recency, based on what Mozilla does. Here's what the results look like on a 15 minute old db backup; first image is sorted by latest, i.e. the old version, second image is trending, i.e. the new version: (note that being ~as good as the old version is good enough for the algorithm since we really just don't want people bumping shitty threads themselves)

**Figma**

screen shot 2018-08-20 at 09 43 26

screen shot 2018-08-20 at 09 50 34

Verdict: Not really a difference in terms of quality at this snapshot in time.

**Codepen**

screen shot 2018-08-20 at 09 43 42

screen shot 2018-08-20 at 09 50 29

Verdict: Algorithmic version seems a little bit better.

**React**

screen shot 2018-08-20 at 09 43 54

screen shot 2018-08-20 at 09 50 24

Verdict: Algorithmic versions looks better.

**styled-components**

screen shot 2018-08-20 at 09 59 25

screen shot 2018-08-20 at 09 59 06

Verdict: Algorithmic versions looks better.

Based on those four hand-picked examples (I didn't check anything else) I say this algorithm looks good for now, let's ship it if the code looks good?

mxstbr:
Just realized my previous migration won't work in production—not sure what to do with all the existing threads just yet...

api/models/thread.js
```diff
@@ -71,14 +71,27 @@ export const getThreadsByChannel = (channelId: string, options: PaginationOption
 };
 
 // prettier-ignore
-export const getThreadsByChannels = (channelIds: Array<string>, options: PaginationOptions): Promise<Array<DBThread>> => {
-  const { first, after } = options
-  
+type GetThreadsByChannelPaginationOptions = {
+  first: number,
+  after: number,
+  sort: 'new' | 'trending'
```
brianlovin, line 77: How do we want these options to be defined and expand over time? Is trending the same as popular? Can we ever sort by date? e.g. trending this week/month/year

api/queries/community/threadConnection.js
```diff
@@ -10,9 +10,15 @@ import {
 import { getThreadsByChannels } from '../../models/thread';
 import { canViewCommunity } from '../../utils/permissions';
 
+export type CommunityThreadConnectionPaginationOptions = {
+  after: string,
+  first: number,
+  sort: 'NEW' | 'TRENDING',
```
brianlovin: Above we had 'new' and 'trending' but here they are all caps. Should we be consistent with these?

api/queries/community/threadConnection.js
```diff
@@ -10,9 +10,15 @@ import {
...
+};
+
 // prettier-ignore
-export default async (root: DBCommunity, args: PaginationOptions, ctx: GraphQLContext) => {
-  const { first = 10, after } = args
+export default async (root: DBCommunity, args: CommunityThreadConnectionPaginationOptions, ctx: GraphQLContext) => {
+  const { first = 10, after, sort = 'NEW' } = args
```
brianlovin: Ditto to caps vs lowercase

api/queries/community/threadConnection.js
mxstbr: No I don't think so? One is the GraphQL stuff (which has to be all caps) and one is internally the argument for the db query. (which we never do all caps)

api/queries/community/threadConnection.js
```diff
@@ -44,10 +50,10 @@ export default async (root: DBCommunity, args: PaginationOptions, ctx: GraphQLCo
     channels = await getPublicChannelsByCommunity(id);
   }
 
-  // $FlowFixMe
   const threads = await getThreadsByChannels(channels, {
     first,
     after: lastThreadIndex,
+    sort: sort === 'NEW' ? 'new' : 'trending',
```
brianlovin: Caps vs lowercase - then we could just say: { sort } if the sort argument is either `new` or `trending`

api/types/Community.js
```diff
@@ -130,6 +130,11 @@ const Community = /* GraphQL */ `
     message: String
   }
 
+  enum CommunityThreadConnectionSort {
+    NEW
```
brianlovin: caps vs lowercase?

api/types/Community.js
```diff
@@ -153,6 +158,7 @@ const Community = /* GraphQL */ `
     threadConnection(
       first: Int = 10
       after: String
+      sort: CommunityThreadConnectionSort = NEW
```
brianlovin: caps vs lowercase?

api/models/thread.js
mxstbr: We'll probably won't be able to do trending this week/month/year with this current setup since we only store the score right now. I think other than that we probably want "TOP", but that's a different calculation altogether.

Not sure, any thoguhts/ideas?

api/queries/community/threadConnection.js
mxstbr: GraphQL stuff

api/queries/community/threadConnection.js
mxstbr: This translates the external GraphQL stuff to the internal db query options.

api/types/Community.js
mxstbr: GraphQL stuff

api/types/Community.js
mxstbr: GraphQL stuff

api/models/thread.js
brianlovin: Not sure either, just trying to think ahead a bit for what other constrains might enter the mix. But also don't want to get too caught up to prevent this from moving forward :)

api/queries/community/threadConnection.js
brianlovin: Referring to this: https://github.com/withspectrum/spectrum/pull/3798/files#diff-8e46094eed25a21b0755138dc52cf17eR77

No worries either way

mxstbr:
Oh damn I thought all our GraphQL enums were purely uppercase, that's why I did the external stuff uppercase and internal stuff lowercase—but I just now realized GraphQL enums can also be lower case 🤦‍♂️ 

Thanks for calling me out on that, you're totally right those should be lower case. On it!

mxstbr:
This should be ready to go, assuming e2e tests pass. The only thing I'm not sure about is how we're going to handle assigning a score to all existing threads—should we just manually add a `calculateThreadScore` job for every thread in the db?

brianlovin:
> should we just manually add a calculateThreadScore job for every thread in the db?

Seems fine to do?

1. Ship mercury
2. Run migration
3. Let mercury process all existing threads
4. Ship the rest

mxstbr:
Shipping mercury and the the API!

PR #3800, opened 2018-08-19
fixed mention regex to respect dots in usernames

dan-weaver:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- desktop

- Bug fix: mention regexp wasn't honoring `.` within usernames

**Related issues**
tangentially related to #3793 #2649 

before this fix:
screen shot 2018-08-18 at 7 59 40 pm

after this fix:
screen shot 2018-08-18 at 7 58 52 pm

### DIFF
--- a/shared/clients/draft-js/mentions-decorator/test/mentions-decorator.test.js
+++ b/shared/clients/draft-js/mentions-decorator/test/mentions-decorator.test.js
@@ -69,7 +69,7 @@ describe('strategy', () => {
       expect(mentions[0]).toEqual('@grdp');
       expect(mentions[1]).toEqual('@rauchg');
     });
-    it('should not include trailing special characters', () => {
+    it('should not include trailing period', () => {
       const text =
         "Hey guys it's @abc.123. where is @xyz? Any news from that guy?  @abc.123... over and out. cc: @some.other.person- @ceo; @hr,.";
 
@@ -77,7 +77,7 @@ describe('strategy', () => {
       expect(mentions[0]).toEqual('@abc.123');
       expect(mentions[1]).toEqual('@xyz');
       expect(mentions[2]).toEqual('@abc.123');
-      expect(mentions[3]).toEqual('@some.other.person');
+      expect(mentions[3]).toEqual('@some.other.person-');
       expect(mentions[4]).toEqual('@ceo');
       expect(mentions[5]).toEqual('@hr');
     });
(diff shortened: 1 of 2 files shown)


mxstbr (changes requested):
Does this tag `Hey @mxstbr.` as "mxstbr." being the mention? That's what I was trying to avoid with not tagging the dot.

dan-weaver:
@mxstbr thanks for pointing that out. I think this fixes it, but I'm not super confident in my regexfoo.  This may also become irrelevant depending on some things in #3793 (assuming the mention plugin is used).

screen shot 2018-08-20 at 10 00 22 am

mxstbr:
Can you try `Hey @dan.weaver... What's up?`? I think that'll still break... I'm going to look at that PR today!

dan-weaver:
Looks like it still works.  I added a test case as well.  Didn't mean to open a can of works though! Trying to think of any other edge cases 🤔 .  Can revisit when and if that other PR is done maybe.
screen shot 2018-08-20 at 10 47 24 am

mxstbr (commented):
Ohh, thank you for adding a test case :+1:

One last question: do we need the `\b`?

dan-weaver:
No it looks like we do not need the `\b`! Removed

mxstbr (commented):
Okay one last nitpick I swear: can we also add a test case for the original edge case (username with a dot in it) in combination with the trailing period one? :angel:

Thank you!!

dan-weaver:
Not a nit pick at all.  This is embarrassing 😊.  I don't think I've had my recommended dose of coffee yet.  The changes to the regex I've introduced allow for other trailing punctuation.  I've updated it again to specify allowed characters instead of disallowed.  this of course means that `_ -` won't be allowed as trailing chars in a mention which I think is a good thing?  They were before though. Let me know if those should be allowed and I can add them.  Also updated the test case.


screen shot 2018-08-20 at 11 26 05 am

mxstbr (commented):
> this of course means that _ - won't be allowed in the mention which I think is a good thing?

I think we want those? Usernames can include `-` or `_`!

dan-weaver:
@mxstbr this latest changes still allows them it just doesn't allow them to be the **last** character of the username.  Which I think is normal for usernames, wasn't sure what your rules were though.
screen shot 2018-08-20 at 11 31 16 am

brianlovin:
> it just doesn't allow them to be the last character of the username

I believe we allow those characters to be the last character of a username

mxstbr (approved):
This LGTM!!! 

Sorry for the slow reply @dan-weaver, thanks so much for the contribution :100:

PR #3806, opened 2018-08-21
New explore communities

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Removed the usage of the curatedContent table from the frontend, instead only rely on the list from the frontent. That means we can much more easily update which communities are featured on the explore page and tweak performance. (it's really slow rn)

Also removed a bunch of communities from explore that aren't really active and highlighted a couple new ones that are very active. (zeit, codepen, next-js)


brianlovin (approved):
LGTM!

Issue #3808, opened 2018-08-21
Thread preview images don't generate for some threads


mxstbr:
This thread breaks the automatically generated thread preview images: https://spectrum.chat/thread/f9cfe7af-e5cb-4830-82a0-4a68e6d0a849

mxstbr:
This is because `window.btoa` cannot handle emojis. Working on replacing our spec-compliant polyfill (`abab` package) with a non-spec compliant polyfill that handles emojis correctly. (`b2a` package)

PR #3809, opened 2018-08-21
Properly guard generating thread meta images

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Related to #3808 (does not fix it though)

This prevents the Twitter preview from looking shit if the thread meta image fails to generate for whatever reason. This is what it looks like right now if it fails to generate:

screen shot 2018-08-21 at 15 53 32

Not very nice lol. This patch makes it so we don't show the `summary_large_image` card if we don't have a large image, basically.

### DIFF
--- a/src/helpers/generate-image-from-text.js
+++ b/src/helpers/generate-image-from-text.js
@@ -38,10 +38,13 @@ type GetMetaImageInput = {
   footer: string,
 };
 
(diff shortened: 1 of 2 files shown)


brianlovin (approved):
Great catch :)

PR #3818, opened 2018-08-22
Fix perpetually being signed out AGAIN

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

The issue with being perpetually signed out turns out to be that the Zeit CDN caches the homepage, and for some reason serves the logged-out homepage even to signed out users! :scream:

[The announcement blogpost](https://zeit.co/blog/now-cdn) said the following:

> Our CDN layer will hit origin every time if it finds it in the Cookie header.

I read that as "any time we find any Cookie header we don't server cached content", but you have to set a specific `__now_no_cache` cookie. :confused: This fixes it by setting said cookie when signing in.

### DIFF
--- a/api/routes/auth/create-signin-routes.js
+++ b/api/routes/auth/create-signin-routes.js
@@ -80,6 +80,11 @@ export const createSigninRoutes = (
         // to the old URL the next time around
         // $FlowIssue
         req.session.redirectUrl = undefined;
+        res.cookie('_now_no_cache', '1', {
+          expires: new Date(2147483647000),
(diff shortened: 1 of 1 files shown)


mxstbr:
Here's the proof is an issue with the CDN one way or another:

screen shot 2018-08-22 at 08 57 25

Session cookie is sent, yet the CDN serves the logged-out homepage back to me via a cache hit.

mxstbr (commented):
Hotfix shipping this

PR #3833, opened 2018-08-24
Always show reputation on user profile

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Previously when a user didn't have a description, we wouldn't show their reputation or website links either—which is kinda strange!

This makes it so we always show those if they have 'em, no matter if they have a description or not.

Ref thread: http://spectrum.chat/thread/67e9d387-2a69-46cb-9745-4aa0b73cad91


brianlovin (approved):
Good catch!

PR #3835, opened 2018-08-24
Fix new messages indicator

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

This thing hasn't shown its face in quite some time due to a broken implementation of `getDerivedStateFromProps`!

This patch fixes it, which means the "New messages" indicator shows again.


brianlovin (approved):
Thank you 👌

PR #3854, opened 2018-08-28
Replace all props.theme function interpolations with static interpolations

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

styled-components has a fast path built-in: when styles don't have any function interpolations, we never recompute them since they can't change.  This leads to big performance gains when re-rendering the same components over and over and over, e.g. on Spectrum when switching between threads in the inbox. (reuses the same components, but with different props)

This patch was done automatically, and moves all of our simple function interpolations in the form of ${props => props.theme.x} to static interpolations that just do ${theme.x}. This allows us to hit the fast path with many more components, making the app re-render faster. This is most noticeable in the dashboard when switching threads.

/cc @uberbryn @brianlovin this would be good to be aware of for any future components

### DIFF
--- a/src/components/badges/style.js
+++ b/src/components/badges/style.js
@@ -1,11 +1,12 @@
 // @flow
+import theme from 'shared/theme';
 import styled from 'styled-components';
 import { Tooltip, Gradient } from '../globals';
 
 export const Span = styled.span`
   display: flex;
-  color: ${({ theme }) => theme.text.reverse};
-  background-color: ${props => props.theme.text.alt};
+  color: ${theme.text.reverse};
+  background-color: ${theme.text.alt};
   text-transform: uppercase;
   padding: 3px 4px;
   margin-left: 4px;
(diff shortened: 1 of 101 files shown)


src/components/badges/style.js
```diff
@@ -21,7 +22,7 @@ export const Span = styled.span`
 `;
 
 export const ProBadge = styled(Span)`
-  background-color: ${props => props.theme.special.default};
+  background-color: ${theme.special.default};
   background-image: ${props =>
     Gradient(props.theme.special.alt, props.theme.special.default)};
```
mxstbr: There's a bunch of cases my automatic move didn't catch, like this one—but it's a good enough start and didn't take a lot of work. 😅

brianlovin (approved):
LGTM!

lachlanjc:
Hey! Sorry to bother you here, but I'd like to do this in my codebases. Is the script you mentioned public?

mxstbr:
No, I hacked it with Atom's find and replace and then some grep and replace magic 🙈

PR #3873, opened 2018-08-31
Enable Apollo Engine again

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api


mxstbr:
Admin shipping so we can start getting some data

PR #3876, opened 2018-08-31
RethinkDB query caching

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [x] Needs testing

- api
- athena
- vulcan
- mercury
- hermes
- chronos
- pluto

----

### README This is based on [this exploration](https://gist.github.com/mxstbr/d1afccc5c094c119eae6fa19ae0a8278), which itself is based on [this](https://gist.github.com/mxstbr/b6c1361bf3208c3d9af488c1e7e7035d) (read the comments) and even earlier [this](https://gist.github.com/mxstbr/4a90621669f6156cb087185b9876ccec).

----

**TODO**

- [x] Implement TagCache
- [x] Implement `createQuery`
- [x] Figure out lifecycles, to e.g. handle `.getAll` single element array responses
- [x] Expire records in redis
- [ ] Move all queries in API to `createQuery`
- [ ] Move API db queries to `shared/` and use them from all workers (otherwise invalidation doesn't work)
- [ ] Set up caching Redis instance that auto-expires after x hours and is throws out LRU
- [ ] Add proper Redis credentials to code so it works in prod


mxstbr:
Idea: we introduce one last optional method called `process` that runs on the result after the query has finished. That was possible before (simple `.run().then(data => /* do something */)`) but isn't possible with this API atm.

This is what e.g. `getUserByUsername` would look like. Since it uses `.getAll` the db query normally returns an array that's either empty or has a single element (`[{ username: 'mxstbr',... }]` or `[]`) but we want to turn it into just the user or null (`{ username: 'mxstbr' }` or `null`):

```JS
const getUserByUsername = createQuery({
  read: (username: string) => db.table('users').getAll(username, { index: 'username' }),
  // This can get the raw data and process it, whatever it returns is returned, thus this should also be what's cached
  // In this case it turns the array with a single element (because we used
  // .getAll it's always an array) into a single element that's maybe null
  process: (raw: Array<DBUser>) => raw[0] || null,
  // This should probably get the processed data? Not sure yet, but this seems to make sense
  tags: () => (user: ?DBUser) => [user && user.id],
});
```

That way the outside API of the query doesn't have to change at all, it stays the same: :100:

```JS
getUserByUsername('mxstbr').then(user => console.log(user.id));
```

mxstbr:
Add the process option and then moved the first two queries over in c5f2891. How's that look @brianlovin? Assuming everything will keep working (🤞 I haven't tested 🙈), can you see any issues with any of our queries that maybe won't work with this?

mxstbr:
Failing e2e test is unrelated, this works*! :tada:

_*at least the queries still work, dunno about the caching_ 😅

mxstbr:
I made the util log the cache hit rate every 30s (even in production) so we can keep an eye out on whether it's helping or not.

In an effort to get some data on whether this is helping at all or not I'd say we move all the user related queries over and move them to `shared/` so they can be used across all workers and start deploying this? Wdyt @brianlovin?

mxstbr:
@brianlovin latest commits overhaul the API again in favor of two separate methods for read and write queries:

```JS
import { createReadQuery, createWriteQuery } from 'shared/rethinkdb/create-query';

const getThreadById = createReadQuery({
  // Note how there's no .run on the end of the query, that makes the caching of the result possible
  query: (threadId: string) => db.table('threads').get(threadId),
  tags: (threadId: string) => (thread: DBThread) => [threadId, thread.channelId, thread.communityId, thread.authorId],
});

const editThread = createWriteQuery({
  // Note, .run() at the end
  query: (threadId: string, data: Object) => db.table('threads').get(threadId).update(data).run(),
  invalidateTags: (threadId: string) => (thread: DBThread) => [threadId, thread.channelId, thread.communityId, thread.authorId],
});
```

This made the implementation in `create-query` much much simpler, it should not be understandable—I'm still not 100% happy with this though. Note how you shouldn't put `.run()` at the end of read queries (that makes caching work since we can key by `query.toString` and store the result from `.run()`), which means I had to put in an extra method to post-process the returned data:

```JS
const getUserByUsername = createReadQuery({
  // Query returns an array...
  query: (username: string) => db.table('users').getAll(username),
  // ...so we process it to only return a single user since username should be unique
  process: (username: string) => (users: ?Array<DBUser>) => (users ? users[0] : null),
  tags: () => () => [],
});
```

In general that's fine, but since the `createWriteQuery` has to be `.run()` it doesn't need the process method since you can just do `.run().then()`, so there's an ugly, slighly unobvious imbalance there. Not sure how to best remedy that yet. 

Wanna take a look at `create-query` and see if you can sorta get what I'm doing there?

mxstbr:
@brianlovin that last commit probably way muddled up the diff (I moved the db.js file to shared), so what you want to be looking at is:

- `api/models/user.js`: see the new query utils in use on a wide variety of read and wide queries
- `shared/db/create-query`: see the implementation of the new query utils

brianlovin:
This feels like solid progress @mxstbr - been studying this for a while this morning and the part I'm not a huge fan of is the tagging methods with functions on functions on functions. Eg:

`tags: () => (user: ?DBUser) => (user ? [user.id] : []),`

I understand why you ended up at this point, but really it would be wonderful to just do:

`tags: [user.id]`

But I understand that means you would have to put the tagging and tag invalidation methods inside query result functions, which makes things also confusing in another way. Do you think there's a way to do better? I feel like every time I write a new query I'll have to re-lookup `shared/db/createQuery` to remember which function takes which arguments in which order.

Eg. 
```
export const getUserById = createReadQuery({
  // ...
  tags: (userId: string) => (user: ?DBUser) => [userId],
});
```

I feel like if I were writing that tag method I'd have to re-lookup every time which function in which order receives which bits of data. Instead, it would be so nice to just write `tags: [userId]`

Maybe it's just a "new" thing right now and we'd get used to it super, quick, too :P

mxstbr:
What if the whole thing was a function so that the input arguments were only passed in a single time? That way you'd only need a single function to get the output of the query, removing that whole double function thing for both `process` and `tags`!

Could look something like this:

```JS
const getThreadById = createReadQuery((threadId: string) => ({
  query: db.table('threads').get(threadId),
  process: (thread: ?DBThread) => thread, // Note: Totally unnecessary in this case, should just be omitted, only here for demonstration purposes of the API idea
  tags: (thread: ?DBThread) => [threadId, thread && thread.communityId, thread && thread.channelId],
});
```

We still have a function for tags and process since you have to get the output somehow but this could be much easier to remember. Wdyt?

> Note: we can't replace the two functions for `process` and `tags` with a single one since the arguments they get are usually not equivalent, those have to stay. If `process` changes the result then `tags` gets the processed data instead of the raw data.

brianlovin:
> We still have a function for tags and process since you have to get the output somehow but this could be much easier to remember. Wdyt?

Yup, totally understand why this is the case. I think I like your solution in the last comment more. Inputs in a more logical place. 

At this point it seems worthwhile to "stress test" this a bit by converting more db queries to see where we hit snags and pain points. I will be able to jump in more deeply here a bit later.

brianlovin:
Why not:

```
getThreadById = (threadId: string) => createReadQuery({ ... })
```

?

I'm probably missing some js thing here that is the obvious answer :P

mxstbr:
Because we'd re-create the query object every time the query is executed—but that's not so bad, we could handle it, I think the bigger problem is that it wouldn't be typed so it wouldn't be enforced to be a fn. I prefer the callback I think

brianlovin:
Fair enough :)

mxstbr:
Latest commits implement the new API idea based on callbacks and moves all existing create-query'ed queries over to that API. Wdyt @brianlovin?

Note, I realized this probably broke db mocking in test, gotta fix that in the future:

- [ ] Make sure test env uses mock db

brianlovin:
This looks + feels so much better :)

mxstbr:
screenshot 2018-10-02 at 15 19 34

Welp fuck

brianlovin:
Well at least you can throw away all those pluto conflicts :)

brianlovin:
Actually almost all of those are just files you can totally delete

mxstbr:
I've now moved all the `api/models/usersNotifications.js` queries over and optimised the shit outta them with two new indexes (there were some slow and buggy ones in there; added `userIdAndNotificationId` and `userIdAndIsSeen` indexes) as I noticed that it's the most-read table by looking at the RethinkDB prod dashboard.

Will move them to shared, and then see if I can share both the user and the usersNotifications queries between all workers and the API!

mxstbr:
Latest commit moves the `user` and `usersNotifications` models from `api/models/` to `shared/db/queries` and rewrites all the imports! Seems like everything is still working 🤞

mxstbr:
Another big one: latest two commits share the `usersNotifications` queries between athena and the API!!! :tada:

Working on sharing these and the `user` queries between all workers, then this should be ready for a first alpha test and then maybe a production deploy! :scream:

mxstbr:
All donezo 💯 User and usersNotifications queries are now shared between all workers and completely rewritten with the cached `createXQuery` utils.

Assuming tests pass, want to give this a thorough review @brianlovin and then let's get this on alpha to see whether something breaks? 👍 (don't merge though, still tons of queries to go)

Also need to run db migrations to add the new indexes to `usersNotifications` I added.

mxstbr:
Looks like three actual test failures, will dig in!

mxstbr:
Okay this should be ready to review and get on alpha @brianlovin. If everything works correctly it'd be ace to a prod test with this limited set of queries before moving on to the rest of them!

brianlovin:
> Okay this should be ready to review and get on alpha @brianlovin. If everything works correctly it'd be ace to a prod test with this limited set of queries before moving on to the rest of them!

Roger that! Would love to take a look a bit later, but don't want to delay you. Will dig in asap :)

api/migrations/20181005143053-users-notifications-useridandnotificationid-index.js
```diff
@@ -0,0 +1,16 @@
+exports.up = function(r, conn) {
+  return r
+    .table('usersNotifications')
+    .indexCreate('userIdAndNotificationId', [
```
brianlovin, line 4: This is fine, but at some point we really need to start scrubbing old notifications. These indexes + the sizes of the original tables are going to blow up our storage volumes. Not too big an issue since it's cheap, but will definitely be something to work on in the future.

api/models/usersChannels.js
```diff
@@ -669,6 +671,29 @@ const getUserUsersChannels = (userId: string) => {
     .run();
 };
 
+const getUserChannelIds = (userId: string) => {
+  return db
+    .table('usersChannels')
+    .getAll(userId, { index: 'userId' })
+    .filter({ isMember: true })
+    .map(rec => rec('channelId'))
+    .run();
```
brianlovin, line 680: Does this then need a `.then(cursor => cursor.toArray())` ? I can't remember exactly when that's needed

api/queries/community/metaData.js
```diff
@@ -1,4 +1,4 @@
-// @flow
+// TODO: Flow type again
```
brianlovin, line 1: Was the flow typing in this file super broken or something? Ideally we don't regress flow coverage, but I understand if it was a nightmare.

shared/db/create-query.js
```diff
@@ -0,0 +1,95 @@
+/* @flow
+ *
+ * Create a RethinkDB query and cache it's result in a TagCache
+ *
+ * Usage:
+ *
+ * const getThreadById = createReadQuery({
```
brianlovin, line 7: I think this documentation needs to be updated with the new callback syntax, right?

shared/db/create-query.js
```diff
@@ -0,0 +1,95 @@
...
+ *   // NOTE: No .run() at the end of the query!!!
+ *   query: (threadId: string) => db.table('threads').get(threadId),
+ *   tags: (threadId: string) => (thread: ?DBThread) => thread ? [thread.id, thread.communityId, thread.channelId] : [],
```
brianlovin, line 10: Same with this

shared/db/create-query.js
```diff
@@ -0,0 +1,95 @@
...
+    const queryString = input.query.toString();
+    const cached = await queryCache.get(queryString);
+    if (cached) {
+      CACHED_RESULTS++;
+      return cached;
+    }
+
+    const result = await input.query
+      .run()
+      .then(input.process ? input.process : res => res);
```
brianlovin, line 68: Does the result need to be explicitly passed to `input.process`?

shared/db/queries/user.js
```diff
@@ -0,0 +1,693 @@
...
+import { removeUsersChannelMemberships } from 'api/models/usersChannels';
+import { disableAllThreadNotificationsForUser } from 'api/models/usersThreads';
+import { disableAllUsersEmailSettings } from 'api/models/usersSettings';
+import { getUserChannelIds } from 'api/models/usersChannels';
+import type { PaginationOptions } from 'api/utils/paginate-arrays';
+import type { DBUser, FileUpload, DBThread } from 'shared/types';
+
+export const getUserById = createReadQuery((userId: string) => ({
+  query: db.table('users').get(userId),
+  tags: (user: ?DBUser) => [userId],
```
brianlovin, line 18: `tags: () => [userId]` right?

shared/db/queries/user.js
```diff
@@ -0,0 +1,693 @@
...
+export const getUserByUsername = createReadQuery((username: string) => ({
+  query: db.table('users').getAll(username, { index: 'username' }),
+  process: (users: ?Array<DBUser>) => (users && users[0]) || null,
+  tags: (user: ?DBUser) => (user ? [user.id] : []),
+}));
+
+export const getUsersByUsername = createReadQuery(
+  (usernames: Array<string>) => ({
+    query: db.table('users').getAll(...usernames, { index: 'username' }),
+    tags: (users: ?Array<DBUser>) => (users ? users.map(({ id }) => id) : []),
```
brianlovin, line 36: I believe this can error if `users` is an empty array. It will throw `cannot destructure property id of null`, or something like that - right? Should it be `users && users.length > 0 ? ...`?

shared/db/queries/user.js
```diff
@@ -0,0 +1,693 @@
...
+  })
+);
+
+export const getUsers = createReadQuery((userIds: Array<string>) => ({
+  query: db.table('users').getAll(...userIds),
+  tags: (users: ?Array<DBUser>) => (users ? users.map(({ id }) => id) : []),
```
brianlovin, line 42: Same here. If `users` is an empty array it will evaluate the left side of the ternary and be unable to destructure `id`. I think...

shared/db/queries/user.js
```diff
@@ -0,0 +1,693 @@
...
+
+          return user;
+        }
+      ),
+    invalidateTags: (user: ?DBUser) => (user ? [user.id] : []),
+  })
+);
+
+export const getUserByIndex = createReadQuery(
+  (indexName: string, indexValue: string) => ({
```
brianlovin, line 113: I realize this is how we have it now in prod, but also realizing that this is a potentially dangerous query if we don't make the `indexName` typesafe to only accept actual indexes we have on this table :P

brianlovin (changes requested):
Left a bunch of individual comments with a couple possible small tweaks. Overall, this looks really nice though. Great work @mxstbr!

api/models/usersChannels.js
mxstbr: rethinkdbdash automatically converts cursors to arrays, only necessary in migrations!

shared/db/create-query.js
mxstbr: No this works as intended! You could also write as `.then(res => input.process ? input.process(res) : res)`, is that more clear?

brianlovin: I guess it is more clear, I didn't realize it would implicitly pass the `res` value to the function. TIL!

brianlovin:
@mxstbr removing the caching just broke all flow coverage for your modified code. I'll see if I can easily remedy.

brianlovin:
https://github.com/withspectrum/spectrum/pull/3876/commits/449aadde10eb974547c80566f09399122a865ad1 is just a temp fix to hush flow on these issues - lmk if that feels okay to you for now @mxstbr :)

brianlovin:
@mxstbr i'm making progress, but most of the static tests are failing with this weird message:

screenshot 2018-10-19 11 49 10

Seems something broke with the schemaGenerator in the last few alpha merges.

brianlovin:
Actually, how could it be that? Static tests pass just fine on our other branches, and this has all the latest alpha commits. Must be something new in this branch specifically breaking the schemaGenerator

brianlovin:
Scratch those last messages. Now it just appears the test_web tests are failing because of mismatched profile/cover photo urls generated for the snapshot tests...weird

brianlovin:
Gonna quickly test if removing the profilePhoto and coverPhoto fields will pass tests. https://github.com/withspectrum/spectrum/pull/3876/commits/a951422ecedd5105b7b235411912583cfc2dc7b2

brianlovin:
Nooooo. Now Peril is failing, but incorrectly so. It seems like it's hitting a rate limit for checking the file's `@flow` declaration, because it definitely exists in the file it's complaining about.

brianlovin (approved):

PR #3884, opened 2018-09-04
Load 25 messages per page in a thread

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

There's really no reason to load 50 messages at once, that's way overboard and blocks slow networks forever. You barely see 10 and we load them in as you scroll pretty aggressively anyway...


brianlovin (approved):
Good call :)

PR #3939, opened 2018-09-21
adds removing username to banning user flow

superbryntendo:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Nothing to deploy. Just a docs update to make sure that banned users' profiles are no longer visible.


brianlovin (approved):
The downside of this is that now someone could re-register a profile with the previous username. Hard tradeoff, I suppose: keep the banned profile visible? Or remove the username and risk someone new swooping the username and having a "bad smell" to their account?

Fine to do it this way for now, as long as we know the tradeoff :)

PR #3944, opened 2018-09-25
Store the full user data in the session cookie

mxstbr:
### Problem

Right now on every single request to the API we do a db query to get the full user data from the userID which is stored in the session cookie.

That's a big issue. Not only does it mean increased pressure on our database, but it slows down every single request you make to the API -even if the code doesn't even need the full user data!

### First Try Solution

My first idea was to replace `req.user` (express) and `context.user` (graphql) with just the user ID and nothing else. In 99% of cases we only use `user.id` and nothing else!

While that would work, those remaining 1% of cases would be timeconsuming to resolve and would require pretty brittle changes. (how can we make sure we've updated everything? we can't)

### Solution

Instead I realized I can just `JSON.stringify` the entire user data to the cookie—that way we just put the entire user data in `req.user` and `context.user` without a db query at all! :tada:

This requires almost no changes and should help reduce pressure and increase performance for every single request you make to the API.

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

### DIFF
--- a/api/authentication.js
+++ b/api/authentication.js
@@ -56,24 +56,25 @@ const init = () => {
   // to avoid having to go to the db on every single request. We have to handle both
   // cases here, as more and more users use Spectrum again we go to the db less and less
   passport.deserializeUser((data, done) => {
-    // Fast path: try to JSON.parse the data
-    // if it works, we got the user data, yay!
+    // Fast path: try to JSON.parse the data if it works, we got the user data, yay!
     try {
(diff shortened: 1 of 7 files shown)


brianlovin (approved):
Seems good to me! Good job digging through all the locations for this change; want to put this on alpha and double check everything before a prod cut?

mxstbr:
Merging to deploy to alpha!

PR #3946, opened 2018-09-26
package.json: remove duplicate version field

ryota-murakami:
**Status**
- [x] Ready for review

I happen to found this, might be missing resolve conflict? Thanks😀


brianlovin (approved):
Good catch!

PR #4007, opened 2018-10-04
Don't re-render desktop navbar on inbox thread change

mxstbr:
This is quite an expensive re-render and totally unnecessary on desktop since the navbar doesn't change at all. Makes switching threads in the inbox feel quite a bit faster!

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)


brianlovin (approved):
Nice catch!

PR #4008, opened 2018-10-04
Fix getDMThread DB query

mxstbr:
.filter can only be used on sequences, not on single records.

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api


brianlovin (approved):
Whoops, thank you for the catch!!

PR #4020, opened 2018-10-05
Deprecate Channel.metaData.threads

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api


brianlovin (approved):
LGTM - technically we still have this field being "queried" from the graphql query itself, but won't take any additional time to resolve so seems fine.

PR #4024, opened 2018-10-05
Publish threads with reaction/messageCount 0

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

Otherwise these threads won't have the denormalized count until somebody replies/reacts


brianlovin (approved):
Whoops!

PR #4025, opened 2018-10-05
Initialize new communities and channels with memberCount: 1

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api


brianlovin (approved):
Facepalm :P

PR #4033, opened 2018-10-06
Use apollo-link-schema for SSR

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Should make for faster HTML rendering as we don't have to go through the network to the API to get the data!


mxstbr:
Here's a deploy: https://spectrum-kfdivlwkkd.now.sh

mxstbr:
Here's how this deploy compares to prod:

screenshot 2018-10-06 at 18 59 45

Not sure what's up with Ohio and Virginia (maybe Now CDN had cached it by the time that was requested?), but overall pretty consistently a fair bit faster.

brianlovin (approved):
Crazy! Ship it :D

PR #4057, opened 2018-10-14
Use em dashes & rich quote

zanedb:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)


brianlovin (approved):
Thanks for the catch!

PR #4080, opened 2018-10-19
Enable CSP

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

**TODO**

- ~~Fix nonce being empty~~ Chrome DevTools just hides it for security reasons 🤦‍♂️ 

Rebase of #2943


mxstbr:
Deploying to alpha, let's see what breaks

mxstbr:
Test failures seem unrelated?

mxstbr:
I think this works! Can you double check me @brianlovin and then ship?

brianlovin (approved):
This LGTM overall, but will be checking more when we have it in prod context so that we can double check it doesn't interfere with user created content.

PR #4146, opened 2018-10-28
Expire cookies after a week, refresh on every request

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)


brianlovin (approved):
I think this works? Can you give it a test on alpha (although the login flow may pose problems with alpha testing...)

PR #4193, opened 2018-11-05
Fix media message previews across the board

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- athena


mxstbr:
For example in DM snippets, before:

screenshot 2018-11-05 at 19 47 39

After:

screenshot 2018-11-05 at 19 47 24

This is pretty much exactly what WhatsApp does. In emails we probably should somehow render the image, but this is a good enough first step that already makes it much nicer than before.

brianlovin (approved):
Good stuff, thank you!

PR #4276, opened 2018-11-20
Force usage of punycode v1.4.1 to remove Node shims from browser bundle

mxstbr:
One of our dependencies was using punycode 2.0.0, but they dropped
support for browsers in v2, causing webpack to include Node browser
shims in our bundle (unnecessarily).

This fixes it by a) setting config.node = false, which should mean this
can never happen again hopefully, and b) force downgrading to punycode
1.4.1

I've also submitted an upstream PR to downgrade, but this is good
enough for now!

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -160,5 +160,6 @@ module.exports = function override(config, env) {
       })
     );
   }
+  config.node = false;
   return rewireStyledComponents(config, env, { ssr: true });
 };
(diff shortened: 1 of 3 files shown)


brianlovin (approved):
Aside from failing e2e, seems good to me!

PR #4297, opened 2018-11-21
Fix posting threads from community view on mobile

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Closes #

When clicking the "Compose" button on mobile on either a community or channel view the right community and/or channel are now pre-selected automatically.

This is done by passing the active community's and active channel's slug via two query params in the URL, e.g.

/new/thread?activeCommunitySlug=spectrum&activeChannelSlug=general


brianlovin (approved):
Excellent find and fix man, this will be super useful

PR #4299, opened 2018-11-21
Upgrade to create-react-app v2 and webpack v4

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Requires us to switch from react-app-rewired (which no longer supports react-scripts v2) to craco, which does support create-react-app v2.

- [x] Make client run again
- [ ] Fix API dev syntax error
- [ ] Make SSR work again


mxstbr:
For some reason the backend servers now throw SyntaxErrors on the flow stuff even though their build process isn't affected 🤔 

Will dig further into this tomorrow, but the frontend side of things seems to work mostly fine!

mxstbr:
Everything seems to build fine again, but the app doesn't rehydrate after SSR for some reason 🤔

PR #4338, opened 2018-11-26
Import specific file from faker

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

Replaces #3722

Speeds up API start time 💪


mxstbr:
Something doesn't work but it's really not worth spending any more time on this, so I'll just close it.

PR #4339, opened 2018-11-26
Don't double-connect to db from workers

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- analytics
- athena
- mercury
- hermes

This was leftover behavior from the move to create-query, but is actually unnecessary.


brianlovin (approved):
Good stuff! Will prod cut this later

PR #4341, opened 2018-11-26
Fix DMThread.snippet database performance

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

Previously this was reading through every. single. message. in every.  single. DM thread. of the user when loading /messages to get the snippets.

You can imagine how performant that was :sweat_smile:

This fixes it by using an existing index to only load the latest message very quickly, reducing reads/sec from (dmthreadamount with a max of 25 * dmthreadmessageamount) to (dmthreadamount with a max of 25)! If you have 25 DM threads with 100 messages each, before you'd be doing at least 2500 reads, now you're doing 25 👍


brianlovin (approved):
Good find!

PR #4343, opened 2018-11-26
fixes issue where scrollbar appears under segmented controls on desktop obscuring the tab indicator

superbryntendo:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)


mxstbr (approved):
Ah damn, my bad, good catch!

PR #4349, opened 2018-11-27
Add communities "memberCount" index, rewrite chronos queries

mxstbr:
Should alleviate some load on the communities table, rather than having to go through all of them this is now mostly instantaneous!

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- chronos

YES


brianlovin (approved):
Great, thank you!

PR #4350, opened 2018-11-27
Use message threadIdAndTimestamp index for getNewMessageCount query

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- chronos

This should reduce load when computing really long threads (> 500 msgs).

### DIFF
--- a/chronos/models/message.js
+++ b/chronos/models/message.js
@@ -31,15 +31,10 @@ export const getNewMessageCount = (
 
   return db
     .table('messages')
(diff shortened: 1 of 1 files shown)


mxstbr:
This is the last somewhat expensive query in Chronos! Once this, #4349 and #4340 land we're good to go with Chronos perf 👍

brianlovin (approved):
<3 <3

PR #4407, opened 2018-12-04
feat: fallback ui for server crashes (500 errors)

### LINKED ISSUE #4403: Fallback UI For Server Crashes
It looks like the server is experiencing some problems at the moment, especially for trying to go to messages. However, the UI is not very user-friendly for displaying these errors. I'm suggesting we work on the UI to ensure that users can still use the web app regardless of errors. For the messages route, this would mean something like keeping the header at the top and maybe doing something like a vertical/horizontal centered user-friendly (actionable) error message. Doesn't have to extravagant. Thoughts?

This was when I tried to go to the messages route:

screen shot 2018-12-03 at 9 20 16 pm

iamclaytonray:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Closes #4403 

screen shot 2018-12-04 at 9 03 21 am


iamclaytonray:
I didn't see tests in `hyperion/` so I did not add any. I'm more than happy to add them. I looked at the contributing section and it didn't look like there were any rules around testing so unless told to, I'll just leave the PR as-is. (Or unless I need to make changes, I'd obviously make the changes)

mxstbr (commented):
This is looking good! Just a small nitpick, let's make the top text "Oops, something went wrong. Sorry!" and then this is good to go 👍 

We don't have a test suite for hyperion right now, so don't worry about that.

mxstbr (approved):
Perfect, thank you so much!

iamclaytonray:
Sure thing! I know the whole github/spectrum thing is underway but if you want a test suite for `hyperion`, I could certainly use the practice. If not, I'll try tackling other existing issues (which is probably smarter to do anyways)

mxstbr:
I'm pretty sure testing hyperion would be pretty hard, so I think you'd be better off tackling other smaller issues to be completely honest! 👍

Issue #4406, opened 2018-12-04
Throttling the community search seems to provide inconsistent intermediate results


raunofreiberg:
https://github.com/withspectrum/spectrum/blob/f57c9f1674a737020feec1d6d9f45c081ed3e1a5/src/views/explore/components/search.js#L61

Searching for `react` provides me with the Reach community [for example](https://streamable.com/m75c8). This does not seem like a good experience, imho. I assume this is because of throttling since the search gets called per 500ms [(btw, the comment is also invalid since it says 200ms, but is actually 500ms)](https://github.com/withspectrum/spectrum/blob/f57c9f1674a737020feec1d6d9f45c081ed3e1a5/src/views/explore/components/search.js#L60). Maybe debounce the search instead of throttling?

mxstbr:
Yeah this should definitely be debounced, not throttled—PR very welcome!

PR #4408, opened 2018-12-04
Debounce communities search query on front page instead of throttling

raunofreiberg:
PR to kick off #4406. 

I started with the most obvious one which I pointed out in the issue. Although, there are a few more searches where debouncing can be considered instead of throttling. These include:

- [Admin communities search](https://github.com/withspectrum/spectrum/blob/181cf90f8d1c49aeb97f0dbb80359daa0a1343a3/admin/src/views/communities/components/search/index.js#L43)
- [Admin users search](https://github.com/withspectrum/spectrum/blob/181cf90f8d1c49aeb97f0dbb80359daa0a1343a3/admin/src/views/users/components/search/index.js#L49)
- [New user onboarding community search](https://github.com/withspectrum/spectrum/blob/eac3de8310f9e683f23a3939e046b9e11b394b4d/src/views/newUserOnboarding/components/communitySearch/index.js#L61)
- [New thread users search](https://github.com/withspectrum/spectrum/blob/83be33bcd95d290a752d5cc9eb0ef8284e4b0359/src/views/directMessages/containers/newThread.js#L120)


mxstbr (changes requested):
This is much better, but now it shows the "Cannot find community for X" while typing. We shouldn't show anything until the search is run, and if the user types anything remove the shown results!

mxstbr (dismissed):
Cleaned this up, thanks so much for the contribution @raunofreiberg! :100:

mxstbr:
Also cleaned up the new user onboarding community search to be debounced 👍

mxstbr:
Latest commit also cleans up the DM thread creation user search to be debounced and properly render a loading state :100:

@brianlovin want to give this a review? :pray:

brianlovin:
@mxstbr this is so much better, awesome work! What would you think of also updating the thread search on a community profile? The placeholder text also doesn't make much sense either, we should follow the pattern that you have in this PR.

screenshot 2019-02-11 17 47 37

mxstbr:
I am going to ship this and then open an issue for the thread search 👍

raunofreiberg:
Nice. Thanks for the cleanup!

mxstbr:
Thanks for contributing!!

PR #4426, opened 2018-12-07
fix(cypress): grab the found user's id

iamclaytonray:
It didn't look like the `getUser` function was ever called and I thought it was odd that the user ID that's returned is the one provided as an argument to the function so I changed this command to grab the found user's ID.

(Unless there was a reason that you guys were doing this.)

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -16,7 +16,7 @@ Cypress.Commands.add('auth', userId => {
 
   return cy.setCookie(
     'session',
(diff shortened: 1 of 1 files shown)


mxstbr (approved):
Aha, good catch! We were calling `getUser` because we needed to put the full user data in the cookie, rather than just the ID, which is no longer the case.

This seems fine though, shouldn't really make a difference 👍 Thank you!

PR #4427, opened 2018-12-07
Implement custom DB connection pool

mxstbr:
**Status**
- [ ] WIP
- [ ] Ready for review
- [x] Needs testing

- api
- hyperion (frontend)
- athena
- vulcan
- mercury
- hermes
- chronos
- analytics

This patch disables the rethinkdbdash connection pool and only uses a single, manual connection for all queries. This is currently live on the prod API for testing, will report back how it goes!


mxstbr:
Update: this is much much better in terms of stability, we haven't crashed in 15 minutes with any connection errors. :tada:

I noticed some periodic slowness of API requests though, which I assume is due to queries being queued while the single connection each API instance has is busy. I figured we do need a little bit of a connection pool. The latest commit implements a custom connection pool on top of rethinkdbdash using the battle-tested and pretty widely used `generic-pool` package.

This is going live in production, let's see how this holds up!

mxstbr:
I'm running into memory leaks almost immediately after startup in production. I assume it must be the generic-pool causing it, as we haven't seen mem leaks in... forever.

Going to try to switch to a different pool package.

mxstbr:
A different pool package with more aggressive timeouts immediately revealed the reason the site is still slow sometimes: establishing new database connections can take >30s, which is obviously very bad. Going to have a think about what could be causing this, it's definitely not load on the database though, I can't correlate high load with slowness (especially not after #4428)

mxstbr:
Running on a single connection is a lot more stable than using a connection pool. I've reverted back to the single connection (+ a connection for changefeeds) on prod and while it's still slow, it's not as horrible as the connection pool version.

Very strange...

mxstbr:
Latest commit switches us to use a fixed list of `n` (currently n = 5) database connections that we rotate between, rather than only a single one. Let's see if that helps performance, it's kind of like "connection pool-lite" 😅

mxstbr:
Still very slow, I'm not sure where to go from here.

PR #4445, opened 2018-12-11
Track database connection count and queue size in DataDog

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)
- athena
- vulcan
- mercury
- hermes
- chronos
- analytics

This is based on #4444 and adds tracking of the database connection count and query queue size over time in DataDog.

I'm not sure we should ship this as it changes _very_ often, many times per sec, which is why I've submitted it as a separate PR. Trying to get input from the observability team on this.

### DIFF
--- a/shared/db/db.js
+++ b/shared/db/db.js
@@ -4,6 +4,7 @@
  */
 import fs from 'fs';
 import path from 'path';
+import { statsd } from '../statsd';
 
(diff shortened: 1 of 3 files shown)


mxstbr:
This should be fine (see convo in Slack), so I'm going to merge this into #4444 and update to use tagging!

PR #4472, opened 2018-12-14
Plaintext markdown chatinput

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

Testing whether switching to a plaintext editing experience could be viable.


mxstbr:
I've rewritten the chat input from scratch with the minimal amount of functionality necessary (and with hooks because why not!). The only thing that works is sending messages with markdown in threads.

And… it seems to be fine and doable! 👍 I estimate we can get the plaintext chat input with full functionality to a shippable state in 1-2 days. The thread editor would take a bit longer, I think we could get that ready to go in a week.

**Pros**

- Consistent browser support, including Android
- Less buggy
- Less complexity

**Cons**

- No WYSIWYG preview (although we could add a preview tab to the thread editor a la GitHub)
- Strange sending UX: first shows raw text from optimistic response, then jumps to formatted once converted on the backend. Might be able to work around this.

I think we should do this given how many complaints we get about the text writing experience. What do you think @brianlovin?

package.json
```diff
@@ -38,7 +38,7 @@
     "prettier": "^1.14.3",
     "raw-loader": "^0.5.1",
     "react-app-rewire-hot-loader": "^1.0.3",
-    "react-hot-loader": "^4.3.11",
+    "react-hot-loader": "^4.6.0",
```
mxstbr, line 41: Had to upgrade to get compatibility with react hooks

package.json
```diff
@@ -145,12 +145,12 @@
...
     "raf": "^3.4.0",
     "raven": "^2.6.4",
-    "react": "16.4.2",
+    "react": "^16.7.0-alpha.2",
     "react-apollo": "^2.3.2",
     "react-app-rewire-styled-components": "^3.0.0",
     "react-app-rewired": "^1.6.2",
     "react-clipboard.js": "^2.0.1",
-    "react-dom": "16.4.2",
+    "react-dom": "^16.7.0-alpha.2",
```
mxstbr, line 156: Had to upgrade to get react hooks

mxstbr (commented):
Note that the meat of the change right now is in `chatInput/index.js` (https://github.com/withspectrum/spectrum/pull/4472/files#diff-f37f8727e3ee11c56ff3a6ae6894c358) but the diff is basically useless since I rewrote everything—it's probably easier to just look at the entire file without the diff!

mxstbr:
I'm going to go ahead and try to get some more functionality in, let's see how this scales.

mxstbr:
Getting there, latest commits add quoted message and media message support. 

**Todos before this is shippable**

- [x] Local storage persistence
- [x] Photo size errors
- [x] Switch to same markdown conversion library on both front- and backend
- [x] Make sure everything works when creating DM threads
- [x] Make sure everything works in DM threads
- [x] Make sure everything works in threads
- [x] Fix code blocks
- [x] Fix flow
- [ ] Fix e2e tests
- [ ] TBD

brianlovin:
> Strange sending UX: first shows raw text from optimistic response, then jumps to formatted once converted on the backend. Might be able to work around this.

We should just transform the plaintext to draftjs in the optimistic response ;)

> I think we should do this given how many complaints we get about the text writing experience. What do you think @brianlovin?

You know I'm all about doing this - let's go for it! Will jump in today and see if I can help.

mxstbr:
> We should just transform the plaintext to draftjs in the optimistic response ;)

I know but this isn't as easy as it sounds without shipping _a lot_ more code on the frontend, which I'd like to avoid.

brianlovin:
> I know but this isn't as easy as it sounds without shipping a lot more code on the frontend, which I'd like to avoid.

We could render the optimistic response at ~70% opacity to indicate it's in a processing state?

mxstbr:
Yeah I think that'd be a good enough solution for now!

mxstbr:
Latest commit makes optimistic responses 0.7 opacity, looks like this:

optimistic-responses

Is that acceptable from a UX perspective? Still a lot of jumping around in this specific case, code blocks are definitely a worst-case scenario for this 😕

brianlovin:
How much does it add to the bundle to transform the plaintext to draftjs in the optimistic response? This outcome feels a bit janky especially on slow networks. This direction feels super good though as far as simplicity of the input!!

Brian

mxstbr:
The one we use on the backend is [32.5kB](https://bundlephobia.com/result?p=markdown-draft-js@1.3.1)—that's a lot. 😕 The underlying issue is that we'd have to ship a markdown parser to the frontend, which is never cheap, even though we don't really even need all of markdown for messages. I'll see if I can find something or whip something up that'll only handle inline styles and code blocks…

brianlovin:
🙏🙏

Brian

mxstbr:
Found the `draft-js-import-markdown` library which is "only" ~8kB min+gzip and does everything we need: https://bundlephobia.com/result?p=draft-js-import-markdown@1.2.1

Latest commit uses that to richly render optimistic responses on the frontend, works great as far as I can tell! Maybe we should switch to using that one on the backend too for compatibility 🤔

brianlovin:
> Latest commit uses that to richly render optimistic responses on the frontend, works great as far as I can tell! Maybe we should switch to using that one on the backend too for compatibility

Not sure, I'll let you make that call. 8kb sucks, but I know we've been meaning to do some bundle cleanup anyways at some point, I'm sure we're shipping something dumb that we could cut to neutralize the 8kb :P

mxstbr:
Geeeez look at this @brianlovin: 2e9d461 

Basically, we need almost no code to do the local storage persistence since 1. nothing can crash anymore now that we only store plain text and 2. React hooks are epic. This also works better than before since it keys by the thread ID in localStorage, meaning if you switch between threads it'll show the cached/uncached content! 😍 

The only thing I removed was the expiry of cached content—since it's now keyed by the thread there's really no reason to expire anymore.

mxstbr:
There is a bug where you can't create a new DM thread with both an image and some text in the message, but that bug is generally around so I won't fix it since it'd require a bit of a backend refactor which I'd like to avoid.

I'll log it in a new issue.

mxstbr:
This is more or less ready for a full-on review and test. Whaddaya say we get this on alpha for a bit to make sure I didn't miss any kinks?

brianlovin:
@mxstbr seems like failing tests need to be wrapped up. I also tested this on mobile and got a server error when sending a message.

brianlovin:
Flow failures seem to be because Flow doesn't have types for hooks yet. Will silence those.

brianlovin:
@mxstbr there still seem to be a few problems with the new chatinput, can try and dig in more today after I work on the site uptime stuff:
- attach a photo to a message, chatinput loses focus
- flow error on using `reader.result` as a message body - expects a string
- after sending a media message, the media preview doesn't go away
- invalid dom properties `hasAttachment` and `networkDisabled` are getting passed down to the textarea, throwing error in console
- can't send direct messages 
screenshot 2018-12-18 10 04 50

Probably some more stuff - let's make sure we write some solid e2e tests here

mxstbr:
Interesting, I tested sending in both existing and new DM threads and it worked perfectly fine. Do you remember the situation around the DM thread you were replying to?

Thanks for the other notes, will dig in today!

- [x] fix focus behaviour
  - [x] focus after photo attachment
- [x] flow error on using reader.result as a message body - expects a string
- [x] after sending a media message, the media preview doesn't go away
- [x] invalid dom properties hasAttachment and networkDisabled are getting passed down to the textarea, throwing error in console
- [x] can't send direct messages
- [x] Fix flash of plain text in optimistic response
- [ ] Fix e2e tests and add some more

mxstbr:
I'm a dummy, when I switched to the new DraftJS conversion library I forgot to add a `convertToRaw` on the server so sending any text message would fail 🤦‍♂️ Fixed now, I think that also fixes your "Can't send DMs" note!

mxstbr:
Tested everything (DM creation, DM reply, thread reply) and it all seems to work fine! 🤞 Want to give this another review?

brianlovin:
Much better! New issues:
- [x] code blocks don't work in messages - just renders an empty message it looks like
- [x] if you try to reply to a message, it should re-focus the chatinput when you click the 'reply' button
- [x] if you quote-reply a message, while simultaneously uploading a photo, it sends the message + photo properly, but does not remove the quote-reply preview from the chat input; it does successfully remove the media preview though
- [x] Network connection states

Quality of life, but not necessary:
- [ ] is it possible to wrap the chatinput in a media-uploader dropzone? that way someone could either click the 'upload photo' icon, or just drag a photo onto the chatinput to load the preview box.

That's all I can find for now - this is coming along super well man!

mxstbr:
>  if you press 'esc' while quote-replying or uploading a photo (or both) it should remove those from the chat input - e.g. reset to only the input with plaintext

I actually consciously removed this behaviour because it was really confusing to me, especially if both a photo and a quote reply are in the input. I wouldn't _expect_ that to happen as a user, but I'm happy to add it back in if you feel strongly about it, it's simple enough 👍 

Thanks for the feedback, great finds! Will dig into those tomorrow

brianlovin:
> I wouldn't expect that to happen as a user, but I'm happy to add it back in if you feel strongly about it, it's simple enough

I don't feel strongly about it, was just thinking if there's a keyboard-only way to clear those things so you don't have to use the mouse and tap the dismiss buttons. Do you have any intuition on what should happen if you press esc in the input field otherwise?

mxstbr:
> Do you have any intuition on what should happen if you press esc in the input field otherwise?

Nothing, just like if you press escape in this GitHub input box, right? 

> was just thinking if there's a keyboard-only way to clear those things so you don't have to use the mouse and tap the dismiss buttons.

If the input is empty, pressing backspace removes them!

brianlovin:
> If the input is empty, pressing backspace removes them!

Oh! Didn't realize this - that's great then :)

mxstbr:
I'm really not sure why code blocks aren't working, it seems to be a bug with the `draft-js-import-markdown` library.

/cc @sstur look at this:

```JS
const raw = convertToRaw(stateFromMarkdown('```\nconst a = "b"\n```'))

raw === {
  "blocks": [
    {
      "key": "9bjeh",
      "text": "",
      "type": "unstyled",
      "depth": 0,
      "inlineStyleRanges": [],
      "entityRanges": [],
      "data": {}
    }
  ],
  "entityMap": {}
}
```

Why is the fenced code block not rendered?

sstur:
Hey @mxstbr. I'll look into that this weekend and push a fix if possible. Thanks for the repro case, should be added to the tests.

[Edit] It might be related to https://github.com/sstur/draft-js-utils/pull/152

brianlovin:
@sstur hey Simon! Any chance you had an opportunity to look into this one?

sstur:
Hey @mxstbr & @brianlovin!

I think this was fixed with https://github.com/sstur/draft-js-utils/pull/98 quite a while back. I just confirmed by running the test suite locally and even console.log'd the output of your test case:
```
convertToRaw(stateFromMarkdown('```\nconst a = "b"\n```'))
```

screen shot 2019-01-02 at 17 56 32

Any chance you're using a version older than v1.2.0?

brianlovin:
> Any chance you're using a version older than v1.2.0?

Entirely possible, I'll take a look. We use a forked version of draft-js as well that @mxstbr set up which may be involved here.

mxstbr:
We are on v1.2.1 across the board, so that's not it. Our forked version of DraftJS also doesn't touch the conversion behaviour at all, so I'm pretty sure that's not it either.

I wonder what's going on here. Maybe we need to downgrade to v1.2.0?

mxstbr:
I'm getting the same behaviour after downgrading to v1.2.0 🤔

mxstbr:
@sstur here's is a replication with vanilla DraftJS and the latest draft-js-import-markdown from npm: https://codesandbox.io/s/my76vn411y Nothing fancy about this setup at all, just this:

```JS
document.getElementById("app").innerText = JSON.stringify(
  convertToRaw(stateFromMarkdown('test\n```\nconst a = "b"\n```')),
  null,
  2
);
```

Maybe the version on npm is out of date with the version on GitHub? It's for sure a bug with the version on npm.

sstur:
@mxstbr, thanks for the repro case!

As you figured out, there was a mismatch between the code on Github and that of npm, so the tests were passing. [Fixed now](https://codesandbox.io/s/l4rv5w6zqz).

I've published v1.2.2 with the latest code.

mxstbr:
Booom @sstur that fixed it! Thanks so much :100:

screen shot 2019-01-08 at 09 05 48

mxstbr:
Last missing bit is fixing message editing (which is why e2e tests are failing), then this is ready for a solid review @brianlovin!

mxstbr:
Latest commit fixes editing of messages in plaintext. We didn't handle that correctly on the backend but we're good now. :tada:

Let's see if e2e tests pass or if I missed something else—assuming I didn't, this is ready for a good in-depth review @brianlovin!

mxstbr:
Here we go this is ready for a really good review and test @brianlovin, let me know which remaining bugs or issues you find and how you like the code!

brianlovin:
1. I updated my packages based on your latest commits, but codeblocks are still disappearing for me:
https://www.dropbox.com/s/l2lgxcioz53y2ea/Kapture%202019-01-08%20at%2012.25.35.gif?dl=0

2. Markdown only parses double line breaks as new paragraphs, so it's collapsing single link breaks:
https://www.dropbox.com/s/2t9yeef89tfyl6p/Kapture%202019-01-08%20at%2012.26.18.gif?dl=0

3. Editing a single-line message is appending an extra line of whitespace in the edit textarea:
https://www.dropbox.com/s/ktm0zy62vxk5uy5/Kapture%202019-01-08%20at%2012.27.23.gif?dl=0

4. We can revert https://github.com/withspectrum/spectrum/pull/4472/commits/f62c8c61ee402ceedf3e4a1b2d9a0fbd7a1e777b if you don't like that behavior, but I changed it so that you can press enter to use line breaks normally while using the edit message input. However, this is causing the chatinput to randomly get focused:
https://www.dropbox.com/s/l4o1tfwpx7m0quf/Kapture%202019-01-08%20at%2012.28.36.gif?dl=0

*ps sorry for the dropbox links - github uploads are failing for me today*

mxstbr:
Great catches, will see how I can remediate some of these!

Code blocks definitely work for me locally, did you try to `rm -rf node_modules && yarn` in both the base directory and the API directory? **EDIT: After looking at the recording, you just need to update the dependencies in the API directory!** The `draft-js-import-markdown` lib on the frontend is correctly converting the code block on the frontend, but you're running the old version in the API, which is why it disappears!

- [x] Single line break paragraphs
- [x] Don't append extra line of whitespace while editing
- [x] Fix enter jumping to chat input while editing

mxstbr:
I honestly don't really know how to solve single line break paragraphs. @sstur how hard would it be to get support for that into draft-js-import/export-markdown?

brianlovin:
> EDIT: After looking at the recording, you just need to update the dependencies in the API directory! The draft-js-import-markdown lib on the frontend is correctly converting the code block on the frontend, but you're running the old version in the API, which is why it disappears!

Whoops! Thanks for the extra details :)

> I honestly don't really know how to solve single line break paragraphs. 

Only short-term workaround I could think of would be to convert from a 'chat input' style input to a 'comment textarea' style; if it looks like a comment box where 'enter' doesn't send the message by default, we could better explain that it behaves the same way as any other markdown document.

sstur:
@brianlovin Those screencaps are super helpful!

@mxstbr
Re: "Markdown only parses double line breaks as new paragraphs, so it's collapsing single line breaks"

I do believe that is the spec-compliant behavior for Markdown (I use the term "spec" loosely ; )

However I do believe the "Github Flavored Markdown" variant changes that behavior. It considers a single line break as a `<br>` and a double line-break as a new `<p>`. (You can inspect this element to see, notice before the "Re" above there's a `<br>`)

I'd be happy to support both behaviors, in fact, it might already be an option baked in to the underlying [markdown parser](https://github.com/chjj/marked) we use.

Edit: I forget we forked that library instead of importing it, can't remember the reason off the top of my head.

Edit 2: The parsing library [does indeed support github flavored markdown including the linebreaks feature](https://marked.js.org/#/USING_ADVANCED.md) however draft-js-import-markdown is not passing those options through. I made a [PR](https://github.com/sstur/draft-js-utils/pull/157) if you want to take a look.

screen shot 2019-01-09 at 10 14 09

mxstbr:
Latest commit makes single line breaks work, so this is ready for another code review @brianlovin!! :tada:

Thanks @sstur for the quick help, appreciated

brianlovin:
Gah, @mxstbr this is SO close!!! Sorry I just found this one: editing a code block doesn't transfer the code block into the edit message input:

https://www.dropbox.com/s/q2iv1z7bb205uv5/Kapture%202019-01-10%20at%2018.24.24.gif?dl=0

mxstbr:
It does! Code blocks in markdown can be written with four spaces in front of the code (like on StackOverflow), only GitHub-flavoured markdown added support for fenced code blocks—so what you're seeing there is a code block.

@sstur how hard would it be to add a GFM mode to `draft-js-export-markdown`?

sstur:
@mxstbr, I think this is not too hard. Probably something like this:
https://github.com/sstur/draft-js-utils/pull/159
(please take a quick moment to sanity check my work)

Such an option will solve your immediate problem, but I'd venture to say there's a larger underlying issue which I'd describe as follows:

Many different markdown representations can produce a single visual output (such as extraneous newlines, different formats for (a) _italic_ (b) *italic* and (a) __bold__  (b) **bold** (need to view this source to see the difference there) as well as code blocks and probably more. Same is true for many markup formats such as HTML where `<b><i>Hi</i></b>` is the same visual representation as `<i><b>Hi</b></i>`.

If I am understanding your implementation, that information is lost in a round trip to DraftJS model and back to markdown.

Maybe this is not a big deal, if your intent is to normalize the user's input into some well-formed canonical representation (like what Prettier does for my code). But, this may seem quirky to the user in the context of composing messages. You might consider doing what Github does, which is store the original text the user typed, either in addition to or in place of the parsed model.

For Github, you can see that by typing some text in a comment here, then going back to edit it. Your original text will always populate the editor, even if you forgot to close a code block. I believe it's the same in chat programs (can verify with Hangouts Chat, although defo don't base your design choices on that product 😂).

Thoughts?

mxstbr:
I generally 100% agree and I'd love to do it that way, but in our situation right now it's not feasible to spend that much time on engineering a perfect editing experience—it just has to be good enough™️!

sstur:
The `gfm: true` option has landed in `draft-js-export-markdown@1.2.2`. Happy Spectruming!

mxstbr:
Perfect, thank you so much @sstur! Latest commit upgrades to that version 💪 

@brianlovin this is ready for another review. Going to branch off of this and start work on the plaintext editor!

brianlovin:
@sstur you are the best, thank you for all your help here! @mxstbr this feels like we're as close as we'll get until we get more people using this in real situations! Let's ship!

mxstbr:
Deploying to alpha!

mxstbr:
Deployed to alpha, feels pretty good 😎 I think we should alias this to prod!

mxstbr:
This is live in production, and I've also posted release notes! :tada: https://spectrum.chat/spectrum/general/spectrum-v2-6-0-a-brand-new-chat-input~84352ff4-9feb-4cce-ab05-575617ad5176

brianlovin:
@mxstbr I saw a handful of logs come in that indicate we might not be properly handling the backend supporting old draftjs messages being sent. Did you see these?

PR #4533, opened 2019-01-10
Upgrade to apollo-server-express@2.4.0-alpha.0

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api

This implements query parsing and validation caching, reference https://github.com/apollographql/apollo-server/pull/2111, and should thusly provide a nice speedup for our frontend's queries!

Since it's an alpha version, we should ship it to alpha.spectrum first, make sure nothing breaks and only then release to production and keep a close eye on the metrics too.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "algoliasearch": "^3.32.0",
     "apollo-local-query": "^0.3.1",
-    "apollo-server-express": "^2.3.1",
+    "apollo-server-express": "2.4.0-alpha.0",
     "apollo-upload-client": "^9.1.0",
     "aws-sdk": "2.200.0",
     "axios": "^0.16.2",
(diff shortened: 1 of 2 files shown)


brianlovin (approved):

brianlovin:
Throw it on alpha! It does  the parsing and caching out of the box? Or do we have to enable that?

mxstbr:
Good question, I'm not 100% sure as the code comments say it's enabled by default, but the tests say it's disabled by default 😅 

/cc @martijnwalraven

mxstbr:
@abernix answered my on Slack, this is enabled by default! Deploying to api.alpha

mxstbr:
After a short, fine trial run on alpha this is live on production. Not seeing any particularly big speedups yet, but it also doesn't hurt:

screen shot 2019-01-11 at 13 23 22

Going to try increasing the cache limit next week and seeing whether that makes for a more noticeable impact!

mxstbr:
Reassessing after a couple days in prod, it seems to have helped @abernix!

p99 response time over:

- Three days **before** deploying this: **250.78ms**
- Three days **after** deploying this: **230.65ms**

👍

PR #4564, opened 2019-01-23
Move thread editor to plaintext

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

After reassessing the importance of this change due to customer feedback I decided to investigate this after all. From the first exploration it seems doable, with some caveats.

**Todo**

- [x] Move to client-side image uploads
- [x] Automatic embeds
- [x] Make sure editing works
  - [x] Make sure users can remove automatic embeds
- [x] Add preview tab to see generated markdown/thread
  - [x] Fix linkifycation in preview
- [x] Fix syntax highlighting in code blocks
- [x] Add mention suggestions based on #4550 
- [x] Styling
  - [x] Add markdown hints from chat input
  - [x] Add overlay when dragging image
  - [x] Fix code block styling
  - [x] Add visible "Upload image" button like the chat input
  - [ ] ~~Add note that drag and drop uploading is supported?~~
  - [ ] ~~Add note about automatic embeds?~~
  - [x] Make sure composer looks right in all instances:
    - [x] Inbox
    - [x] `/new/thread`
- [x] Merge `src/components/composer` with `src/components/threadComposer/composer` (done by @ThomasRoest in #4575)
- [x] Make sure the merged composers look right
  - [x] Community view
  - [x] Channel view


mxstbr:
The hardest bits to figure out here are going to be images and embeds.

I think images we have to do GitHub-style "upload from client then input markdown link". /cc @brianlovin how would we best accomplish that? New `uploadImage` mutation?

I have no idea how to do the embeds, I think the choice is either to remove that functionality (while keeping existing embeds around, of course) or automatically "unfurling" a list of supported links into embeds (youtube, simplecast, soundcloud, etc) on the backend. Any thoughts?

ThomasRoest:
What are the problems with the current editor?

mxstbr:
It is very buggy, lots of users constantly complain about it (see [my recent blog post](https://mxstbr.com/thoughts/tech-choice-regrets-at-spectrum/)). We want to replace the WYSIWYG part with a plaintext markdown editor, which is then converted on the backend! (this should be mostly irrelevant for your work in #4575 I think?)

brianlovin:
> I think images we have to do GitHub-style "upload from client then input markdown link". /cc @brianlovin how would we best accomplish that? New uploadImage mutation?

Yeah, it might be worth poking around to see if other libs exist for this yet, otherwise we could try and figure out how GitHub handles this by just inserting placeholder markdown while waiting for the upload to return a url. The downside here would be that you create an upload every time, regardless of if the resulting url actually gets posted. But storage is cheap, so not a problem really right now.

> I have no idea how to do the embeds, I think the choice is either to remove that functionality (while keeping existing embeds around, of course) or automatically "unfurling" a list of supported links into embeds (youtube, simplecast, soundcloud, etc) on the backend. Any thoughts?

I'm fine with removing functionality and getting really smart with link unfurling for popular services. I'd add Figma embeds and code sandbox urls to your list. I don't know how that would look in the backend (just inserting an iframe into the draftjs body?) but this feels like the best user experience to me personally.

mxstbr:
The latest commit starts working on the client-side image uploads by adding an `uploadImage` mutation—doesn't quite work yet, but it'll get there I think.

The embed stuff sounds good to me too, inserting an embed into the DraftJS content seems simple enough to be feasible. I'll try making that happen after making sure the image uploads will work.

mxstbr:
I have been battling around for hours with this image uploads stuff now, and can't seem to get it to work. When dropping an image on the new composer, `(await file).stream` always results in an undefined error on the server. Have you ever seen that @brianlovin, any hunches why that could be happening?

brianlovin:
@mxstbr could you push up your latest work? Not sure what could be going on there. Do you have `FILE_STORAGE=s3` set in your env? e.g. `yarn run dev:api:s3` to ensure that uploads are hitting s3 and not your local directory?

mxstbr:
That is not the issue, the `image` argument to the `uploadImage` mutation is just `null` even though it is provided from the frontend 😕 Latest commits are up!

mxstbr:
Oh my god are you kidding me, did that fix it @brianlovin? 🤦‍♂️🤦‍♂️🤦‍♂️  I am such a dummy, sorry about all the struggles here. Cannot believe I spent three days messing around with this.

mxstbr:
😍😍😍😍😍😍

editor-image-uploads

mxstbr:
Latest commits add automatic embeds!!! :tada:

auto-embeds

Currently supported services are the same as before: Figma, Framer, Youtube, Vimeo, Codepen and Codesandbox, as well as support for generic `<iframe>`s.

Note that this only works for the initial publish, automatic embeds are not supported for edited threads as that would add a ton of complexity around edge cases. What happens if the user removes an automatic embed? What happens if they add the same link again? etc. etc.

brianlovin:
Amazing progress here @mxstbr! How does it work if a thread has multiple youtube links? Do those embeds show up a line beneath where the link was? Or at the bottom of the post?

mxstbr:
Yes, they are all added as blocks beneath the paragraph the link was in. I thought that was the nicest behaviour, adding them to the end of a long post might mean they are never used since interested folks would already have clicked on the link, right?

Open to discussing about the fine details of this UX of course, we also need to add a button for image uploads still.

brianlovin:
> I thought that was the nicest behaviour, adding them to the end of a long post might mean they are never used since interested folks would already have clicked on the link, right?

Agreed, this sounds great. I could envision a situation where a person pastes several youtube links and is annoyed by all of them unfurling, but we can get to that later.

mxstbr:
They can remove the embeds that annoy them in the editing mode :+1:

mxstbr:
Latest commit adds a preview tab to the editor, which also previews the automatic embeds!

thread-preview

The automatic linkifycation does not seem to work in the preview and I am not sure why. 🤔 But other than that, this is getting somewhere!

mxstbr:
Now we also got mention suggestions :party_parrot:

thread-mention-suggestions

This is getting fairly close, apart from needing a polish pass over the styling. @brianlovin do you want to look at that?

----

Other than styling polish, all we gotta do is fix the e2e tests and look into the preview not linkifying! :tada: If I can't figure it out quickly let's still ship this once it has the style cleanup?

mxstbr:
We should probably also add the markdown hints from the chat input to the composer finally! @brianlovin want to tackle that too?

brianlovin:
This is awesome @mxstbr - will push up the markdown hints and some styling polish.

Couple broken things in addition to links not rendering in preview:
- can you double check the image paths are working properly - right now images don't render in preview
- code blocks do a weird double nesting:
screenshot 2019-01-31 13 55 39

mxstbr:
I cannot seem to get uploaded images to show up at all, not in the preview and not in the published thread either. It might be related to the local file upload, but I cannot get uploading-to-s3 working either 🙄 The image upload is working fine (if I manually go to `http://localhost:3000/uploads/whatever` it works), but for some reason DraftJS is not rendering them as images to the DOM.

brianlovin:
> but for some reason DraftJS is not rendering them as images to the DOM.

Yikes. Really strange indeed, let me know if I can help poke around here. The only thing that came to mind for me was: are you signing those images properly with the imgix signatures?

mxstbr:
Huh no I am not at all. Do I need to do that manually when calling `uploadImage` from the API?

brianlovin:
You call it at query time - see `api/queries/thread/content.js`

mxstbr:
That turns out not to be the problem. For some reason, no `draft-js-plugins` are applied in the preview, causing images not to show up and links not be linkified. :confused:

Digging into how we can work around this, very annoying.

mxstbr:
What I do not understand is why embeds show up in the preview, but images do not. That makes no sense. Why would the `draft-js-embed-plugin` work but the `draft-js-image-plugin` and `draft-js-linkify-plugin` not work?!

mxstbr:
After digging into this some more, this seems to be related to this issue in `draft-js-import-markdown`: https://github.com/sstur/draft-js-utils/issues/147

Essentially, it does not play well with the `draft-js-image-plugin` right now. While that does not explain the missing linkifycation, it is a start! I am going to dig into how we could resolve this

mxstbr:
Found the fix!! https://github.com/sstur/draft-js-utils/pull/164 Now to wait for @sstur to publish a release and this will be much closer to ready

sstur:
I just cut a release draft-js-import-markdown@1.3.1

mxstbr:
Latest commits fix images in the thread content preview and persisting drafts to localstorage! This is getting very close, going to work on the styling stuff now (adding an "Upload image" button, making sure it is obvious where images can be dropped,...) and then this is ready for another round of reviews @brianlovin

mxstbr:
Updates 😍 

- Much nicer drag-and-drop image upload by inserting the text at the cursor position and replacing selected text:

spectrum-image-upload-replace

- Working image previews:

spectrum-thread-image-preview

mxstbr:
Latest commit adds an overlay when dragging images 😍😍😍 

spectrum-image-drag-overlay

brianlovin:
@mxstbr this is all amazing :) Want to take a stab at putting an 'upload image' button somewhere where you think it makes sense? Happy to jump in after that and adjust styling or visuals

mxstbr:
Done in the latest commit @brianlovin, at least on desktop:

screenshot 2019-02-07 at 13 18 36

This is what the bottom area looks like on mobile:

screenshot 2019-02-07 at 13 19 11

This is obviously not ideal, but we really need the "Upload Image" button on mobile since you cannot drag and drop images. Any better ideas?

----

I also removed the markdown hints in favour of putting the markdown icon there and linking to the GitHub markdown guide. The hints made it seem like you could only use inline styles, but headings and lists and stuff are also supported!

mxstbr:
Found a better placement for that button on mobile!

screen shot 2019-02-07 at 15 54 53

@brianlovin this is ready for another review, and then I have to merge this into #4575 before shipping both of them together 👍

mxstbr:
I just merged #4575 by @ThomasRoest into this PR, meaning we now also have a unified thread composition experience in the slider! :tada:

Gotta clean up some small visual bugs, fix tests and types and then we should be good for another round of reviews @brianlovin

mxstbr:
**Todos**

- [x] Fix overflow on `/new/thread` on desktop
- [x] Fix styling issues on `/new/thread` on mobile 
- [x] Fix flow
- [ ] Fix tests

ThomasRoest:
this looks great! 

composer

brianlovin:
@mxstbr @ThomasRoest this PR seriously makes me so happy. What a huge, massive, seriously big quality of life improvement to composition on Spectrum. I'm running this locally and just so stoked to ship this. Going to do a bunch of testing and make sure we can't suss out some lingering bugs.

brianlovin:
@mxstbr found some bugs + notes:

- [x] links don't actually work. type `[link](https://google.com)` and it renders as a link, but the element isn't receiving the `href` attribute
- [ ]  how would you feel about also moving the edit experience to this modal? might reduce a ton of complexity from the edit screen/thread view; i'm a bit worried that the edit view has the old draftjs rich preview editor, which might introduce bugs cc @ThomasRoest 
- [x] in the preview sometimes images have this draftjs focused state:
screenshot 2019-02-08 14 04 20

  - [x] if you publish a thread with an image in this focused state, the final rendered thread will keep the focused styling

mxstbr:
> how would you guys feel about also moving the edit experience to this modal? 

The "convert draftjs to markdown for editing" thing that we do for messages won't work for the complex styling of threads. For example, what do we do with automatic embeds?

If we do move editing to plain text we would need to store the raw markdown body on the server and send that down the wire. 

I think that is more work than necessary for this v1, I would rather ship this and then tackle plaintext thread editing in a followup?

----

Noted on your other points, good catches! Will investigate on Monday

ThomasRoest:
see `components/rich-text-editor/Image.js` for styling the image overlay / focus state in ImageContainer and ActiveOverlay

mxstbr:
Yeah I think we can just entirely get rid of those focus states since they were only relevant during WYSIWYG editing. Want to submit a PR against this branch @ThomasRoest?

mxstbr:
After thinking about this more, really the only thing the plaintext input does not account for is the automatic embeds. If we are fine with users not being able to remove them, we can move the editing experience to plaintext too.

That being said, it would require a lot more work to make plaintext editing work on both the back- and the frontend, so I would much rather ship this for now and then work on that afterwards if we are cool with that @brianlovin?

brianlovin:
> That being said, it would require a lot more work to make plaintext editing work on both the back- and the frontend, so I would much rather ship this for now and then work on that afterwards if we are cool with that @brianlovin?

Okay, I understand that's challenging. Since this is still a huge incremental improvement, let's 🚢 and figure out editing down the road. My 2 cents though is that I'm not worried about the user being able to edit the embeds after the fact.

brianlovin (approved):

mxstbr:
Merging!!!!! :scream: :tada: :tada: :tada:

Will get this live on alpha.

Issue #4556, opened 2019-01-19
Make inbox filters links


mxstbr:
- [x] Feature request

https://github.com/withspectrum/spectrum/blob/9bb7010cd0f22d4a59e1a74f185cdc4a78ef1b28/src/views/dashboard/components/communityList.js#L110-L139

This whole bit should be a link that redirects to `/communityslug` when the JS hasn't rehydrated.

srikar0896:
@mxstbr can I pick this up?

mxstbr:
Yes please @srikar0896!

srikar0896:
Hi @mxstbr, It's my first pr, I am trying to get aware of codebase. It's a bit unclear. What do you mean by redirect to `/communityslug` **when the JS hasn't rehydrated**? What is be the expected behaviour?

srikar0896:
Okay, I want to get some context here. So If we are simply redirecting the user to `https://spectrum.chat/${community.slug}` onclick why do we need `onClick` handler. So do we want to simply redirect the user to `https://spectrum.chat/${community.slug}` onClick or do something and then redirect in `onClick` handler.

mxstbr:
We only want the user to redirect if the JavaScript has not rehydrated yet. Otherwise, if you click on it, nothing happens.

When the JS has rehydrated, we do not want to redirect, we want to filter the inbox by that community, so we let the `onClick` handler do its thing!

Does that make sense? Sorry for not explaining it well :pray:

mxstbr:
> no matter what it does in onClick handler it always redirects to /communitySlug right.

I think this is not true—the `onClick` handler only needs to call `evt.preventDefault()` and the `href` will not be opened I'm pretty sure!

So we don't need to know whether JS has rehydrated or not, we only need to call `evt.preventDefault()` from the `onClick` and we should be good.

srikar0896:
@mxstbr yes, that makes sense. But If we wrap the `communityListItem` with `Link` we don't even need to call `e.preventDefault()` in `onClick` handler. Please review https://github.com/withspectrum/spectrum/pull/4565

PR #4565, opened 2019-01-23
Clicking on the community in the sidebar must redirect the user to community home page.

srikar0896:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

**Deploy after merge**
- hyperion (frontend)

**Release notes for users**
- User can visit the community home by clicking the community in the sidebar.

**Related issues**
Closes https://github.com/withspectrum/spectrum/issues/4556

### DIFF
--- a/src/views/dashboard/components/communityList.js
+++ b/src/views/dashboard/components/communityList.js
@@ -107,36 +107,40 @@ class CommunityList extends React.Component<Props> {
           </CommunityListItem>
           {sortedCommunities.map(c => (
             <ErrorBoundary fallbackComponent={null} key={c.id}>
-              <CommunityListItem
-                onClick={() => this.handleOnClick(c.id)}
-                active={c.id === activeCommunity}
-              >
-                <CommunityAvatarContainer>
-                  <CommunityAvatar community={c} showHoverProfile={false} />
-                </CommunityAvatarContainer>
-                <CommunityListMeta>
-                  <CommunityListName>{c.name}</CommunityListName>
-                  <Reputation
-                    ignoreClick
-                    size={'mini'}
-                    tipText={`Your rep in ${c.name}`}
-                    reputation={c.communityPermissions.reputation}
-                  />
-                </CommunityListMeta>
-
-                {c.id === activeCommunity && (
-                  <ErrorBoundary>
-                    <SidebarChannels
-                      activeChannel={activeChannel}
-                      communitySlug={c.slug}
-                      permissions={c.communityPermissions}
-                      slug={c.slug}
-                      id={c.id}
-                      setActiveChannelObject={this.props.setActiveChannelObject}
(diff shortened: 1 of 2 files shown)


mxstbr:
Deployed as https://spectrum-ejrmdorrgw.now.sh for testing, let's see...

src/views/dashboard/components/communityList.js
```diff
@@ -107,36 +107,40 @@ class CommunityList extends React.Component<Props> {
...
-                {c.id === activeCommunity && (
-                  <ErrorBoundary>
-                    <SidebarChannels
-                      activeChannel={activeChannel}
-                      communitySlug={c.slug}
-                      permissions={c.communityPermissions}
-                      slug={c.slug}
-                      id={c.id}
-                      setActiveChannelObject={this.props.setActiveChannelObject}
+              <Link to={`/${c.slug}`}>
```
mxstbr: We do not want to redirect to the community once the JS has rehydrated. I think to fix that, we need to move the `onClick` to the `Link` here!

```suggestion
              <Link to={`/${c.slug}`} onClick={() => this.handleOnClick(c.id)}>
```

src/views/dashboard/components/communityList.js
```diff
@@ -107,36 +107,40 @@ class CommunityList extends React.Component<Props> {
...
+                <CommunityListItem
+                  onClick={() => this.handleOnClick(c.id)}
```
mxstbr: And remove the onClick here:

```suggestion

```

mxstbr (changes requested):

src/views/dashboard/components/communityList.js
srikar0896: oh yes. And adding `e.preventDefault` in `onClick` handler might satisfy that condition.

Thanks.

mxstbr:
Hmm, this still redirects me to `/${communitySlug}` when clicking on a filter, even when the JS is running?

srikar0896:
@mxstbr yes, the event in `sideBarChannel` is stopped from propagating to the parent but the default event for a link is to redirect to that link, so I have fixed that by adding `e.preventDeault()` because we don't want the default action tobe taken.

srikar0896:
@mxstbr Hi!! Have you got a chance to check this?

Ponjimon:
I had a look at this and unfortunately, it does not work for me. I mean, it does, but I'm not sure if that's intended. If I click the community item, it just redirects me to the community instead of expanding the channel list of that community as it currently does.

If you put the `e.preventDefault()` at the beginning of the onClick function to capture the regular behavior of a link, then it will work as it should, in my opinion. The current behavior is kept but additionally, it's not a link that you can open in a new tab. No need for the `else` here, we can always prevent the default.

PR #4576, opened 2019-01-24
Fix hyperion memory leak

### LINKED ISSUE #4573: Fix hyperion styled-components memory leak
We need to manually call `sheet.complete()` if an error happens, see 
https://github.com/styled-components/styled-components/issues/1624#issuecomment-425382979

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Closes #4573

@brianlovin this might be causing the response time jumps every couple days? Not sure, but it should definitely alleviate some memory pressure on our hyperion instances.

### DIFF
--- a/hyperion/renderer/index.js
+++ b/hyperion/renderer/index.js
@@ -189,6 +189,8 @@ const renderer = (req: express$Request, res: express$Response) => {
       );
     })
     .catch(err => {
+      // Avoid memory leaks, see https://github.com/styled-components/styled-components/issues/1624#issuecomment-425382979
+      sheet.complete();
       console.error(err);
       const sentryId =
         process.env.NODE_ENV === 'production'


hyperion/renderer/index.js
```diff
@@ -189,6 +189,8 @@ const renderer = (req: express$Request, res: express$Response) => {
       );
     })
     .catch(err => {
+      // Avoid memory leaks, see https://github.com/styled-components/styled-components/issues/1624#issuecomment-425382979
+      sheet.complete();
```
lucasfeliciano, line 193: This api has changed.

Currently is `sheet.seal()` instead of `sheet.complete()`

Also would be even better if you call it on the `finally` block instead of catch. :)

lucasfeliciano:
> this might be causing the response time jumps every couple days? Not sure, but it should definitely alleviate some memory pressure on our hyperion instances.

This was our case as well, the kubernetes pod would just stop responding and restart at some point and some requests would just stay in a limbo state.

Just to give you some idea about this change:

image

You can see how fest the memory would increase in the beginning of the graph.
When it goes down  means that a deploy happened.
And after the last deploy you see that the memory increase is much smoother

hyperion/renderer/index.js
mxstbr: We're on v3, so it's `.complete` right?

mxstbr:
@lucasfeliciano would you mind updating the styled-components documentation to include this information? :pray: Otherwise everybody's going to run into this issue!

hyperion/renderer/index.js
lucasfeliciano: I'm 100% sure when the API changed, we are on v6 so it might be that you need to call `.complete` instead

brianlovin (approved):
Ship it!

PR #4647, opened 2019-02-15
Add request-level rate limiting

### LINKED ISSUE #2776: Implement rate limiting for our Node servers

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

- Based on #2789: this is a copy-and-paste of @LetItRock's great implementation with a tiny bit of code cleanup :+1:
- Closes #2776

Stores rate limits in redis. The chosen limits are:

- API: 40 req / 10s to allow for short bursts of API requests during initial page loads
- Hyperion: 5 req / 20s to make sure we never go above the 1 rps limit our servers can handle (users should also really never load more than one page, as afterwards everything is client-side rendered)

We will have to monitor 429 responses after deploying this to ensure these limits are not too low. (esp the API) We track those in DataDog, so it is simply a matter of creating a graph for them.


brianlovin (approved):
LGTM but it'd be great if you could deploy and monitor this during your daytime in case the limits need adjustments or anything breaks :)

mxstbr:
Deploying and monitoring!

PR #4664, opened 2019-02-18
fix bug: cannot upload larger than 500k cover photo when save community settings

tuoxiansp:
image

### DIFF
--- a/src/views/communitySettings/components/editForm.js
+++ b/src/views/communitySettings/components/editForm.js
@@ -191,7 +191,6 @@ class EditForm extends React.Component<Props, State> {
       website,
       file,
       coverFile,
-      coverPhoto,
       communityId,
       photoSizeError,
     } = this.state;
@@ -201,7 +200,6 @@ class EditForm extends React.Component<Props, State> {
       website,
       file,
       coverFile,
-      coverPhoto,
       communityId,
     };
 


brianlovin:
@tuoxiansp I'm not totally sure I understand this PR. Could you please describe the bug in more detail and why these changes fix the bug?

tuoxiansp:
image

Users can upload there community cover photo that is less than 3mb .

But actually they'll get failed when uploading more than 500kb . 

image

This is the request payload of updating community cover .

'coverPhoto' was a huge base64 string in this payload . 

This cause body-parser down.

body-parser was limited 100k by default.

And we can't turn up the limit like this:

```javascript
bodyParse.json({limit: '3mb'})
```

Cause though we can successfully upload large images, But images will be base64 encode in db .

That makes website slow with api response size can up to 8 mb or larger.

That's the problem.

----

And what I'm doing:

I'm not really know what happening, just diff the userSettings/components/editForm and communitySettings/components/editForm, found the right way is not to input 'coverPhoto'.

image

tuoxiansp:
gif: https://static.ucloud.cn/002273033c9349fdb76bd2f7fd36afd9.gif 

Check this page: https://spectrum.chat/b1ncer-test/settings

You can find "getCommunityChannels" returned a base64 encoded image string on key 'coverPhoto'.

tuoxiansp:
#4711

mxstbr:
@brianlovin I think this could make sense, right now we are sending both `coverFile` _and_ `coverPhoto` which is apparently too much?!

brianlovin:
Yep, this works locally and can reproduce the failed coverphoto upload locally as well on `alpha`. Thanks so much for the find here @tuoxiansp

brianlovin (approved):

PR #4737, opened 2019-02-26
Add shift key listener when deleting messages

Ponjimon:
This allows skipping of the confirmation modal when pressing shift while clicking the message delete button. This allows faster moderation. I got this idea from Discord (they're doing that too).

Thought it would be a nice little feature. If not wanted, just reject it.

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

- The delete message button can now be clicked while pressing the Shift key to skip confirmation

### DIFF
--- a/src/components/message/index.js
+++ b/src/components/message/index.js
@@ -44,6 +44,8 @@ import {
 } from './style';
 import getThreadLink from 'src/helpers/get-thread-link';
 import type { GetThreadType } from 'shared/graphql/queries/thread/getThread';
+import deleteMessage from 'shared/graphql/mutations/message/deleteMessage';
+import { deleteMessageWithToast } from 'src/components/modals/DeleteDoubleCheckModal';
 
 type Props = {|
   me: boolean,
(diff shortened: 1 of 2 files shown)


brianlovin:
This seems like a good idea, I'm fine with it! Looks like just some static type failures there in CI - want to snag those @lookapanda and we can do a last test + merge?

Ponjimon:
Will look into it tomorrow!

mxstbr (commented):
I like it! 👍 Maybe that should be `CMD` instead of Shift?

Ponjimon:
Why should it be the `CMD` key instead? Windows does not have that key btw, both systems have Shift though.

brianlovin:
> Why should it be the `CMD` key instead? Windows does not have that key btw, both systems have Shift though.

I don't have a preference of cmd vs shift, but I think Max is just referring to the system modifier key here, so on Windows it'd be the windows key

Ponjimon:
I fixed the type failures, but what should I do with the key? Should I leave the Shift key or change it to support the system modifier key?

Ponjimon:
I also just noticed this: https://stackoverflow.com/a/3922353/2632061

According to this, the `CMD` key is not a modifier key and browser-dependant. So for me, it sounds like a better option to stick with Shift. 

Of course, correct me if I'm wrong.

mxstbr:
I am fine with shift if that is the existing convention to skip confirmation dialogs! 👍

brianlovin (approved):

PR #4772, opened 2019-03-01
Move markdown conversion to microservice

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

In the end we should no longer be shipping any DraftJS or ImmutableJS code, so we need to move all instances of markdown => DraftJS or DraftJS => markdown to the new convert.spectrum.chat microservice.


mxstbr:
The hardest one here is definitely going to be the optimistic message update. Not sure how to do that yet 😕

mxstbr:
The latest commit makes it so we convert to DraftJS on the client via the network and show that converted state in the optimistic update.

This would mean we can kick DraftJS and ImmutableJS out of the bundle and the UX is fine, but it also means that it takes a couple hundred milliseconds until you see your message after pressing "Enter".

@brianlovin can you comment on the UX tradeoff here? We have a smaller JS bundle (i.e. faster pageloads and better perf across the board) on the one side, and slightly slower message sending on the other. 

The other option I can think of would be to show the raw markdown in the optimistic update, and then replace it with the converted version once the API responds. We had that initially but you did not like it as much.

Any thoughts?

brianlovin:
What if we used a tiny tiny markdown parser like https://github.com/developit/snarkdown to show the optimistic update, which gets overwritten with the draftjs stuff with all the formatting and embed logic? That way it's slightly less jarring with smaller impact on the bundle.

mxstbr:
Good idea, I will have a look around for something! The problem with snarkdown is that it returns a HTML string, which is not what we want 😕

mxstbr:
I made it happen in the latest commit! :tada: :tada: :tada:

Back on track to finish this soon hopefully 🔜

mxstbr:
The latest commit removes all DraftJS code from our frontend bundle!!! 😍🎉🎉🎉 All conversion on the frontend now happens via `convert.spectrum.chat`.

File size differences:

```
425.71 KB (-65.27 KB)   main.ed466c85.js
32.4 KB   (-858 B)      Thread.a7b4f27f.chunk.js
30.23 KB  (-890 B)      DirectMessages.98bcd9b1.chunk.js
27.3 KB                 Splash.1a360120.chunk.js
23.21 KB  (-589 B)      communitySettings.107661c7.chunk.js
17.6 KB                 NewCommunity.f04879f2.chunk.js
16.55 KB                Dashboard.b5d4c8d5.chunk.js
16.44 KB                ChannelView.4bae12a9.chunk.js
15.18 KB                channelSettings.54cfa04d.chunk.js
14.74 KB                CommunityView.092aaa64.chunk.js
13.72 KB                Search.822780ff.chunk.js
11.79 KB                UserView.514d939d.chunk.js
8.98 KB                 UserSettings.070fcbf4.chunk.js
7.56 KB                 sw.js
6.73 KB                 Explore.d5da766c.chunk.js
3.31 KB                 Notifications.5b4eac6c.chunk.js
1.04 KB                 bootstrap.js
```

I think we can get to a point where we do not even _depend_ on DraftJS anymore entirely, and only operate on the raw content state. Will see what would be necessary to get there!

mxstbr:
Well, that was maybe overly optimistic—we are quite a ways away from completely removing DraftJS as a dependency, but this is a good start.

@brianlovin this is ready for a review! Make sure to double check the UX in all the conversion cases: previewing/editing threads, optimistic message sending, editing messages. Does it all look fine to you?

brianlovin (changes requested):
This LGTM so far, I think the loading states are just a bit jarring. Can we add a timeout for the message editing and thread editing states? 

Video: https://www.dropbox.com/s/1gu9q8ki9uytf8s/markdown.mov?dl=0

Also in that video it looks like extra lines are getting appended to messages.

mxstbr:
Fixed the lines getting appended in #4848, working on delaying loading states!

mxstbr:
Looking at Apex Ping, a network request to the markdown converter takes a mean time of 266ms:

Screenshot 2019-03-09 at 13 44 04

Will play around with the delay value so we only show the loading indicator once necessary!

brianlovin:
That's quite a long time, don't you think? I mean, it's fast, but I suppose I would have assumed it would be less than 100ms.

mxstbr:
I tried adding a delay to the message editing loading state in 4ec1af9b2, but that actually felt even worse. Instead, I added a nicer loading state that does not cause things to jump around after clicking "Edit" and it's much better:

Kapture 2019-03-09 at 13 59 11

This is ready for another review 👍

mxstbr:
> That's quite a long time, don't you think?

🤷‍♂️ We can optimise that down the line to not do the conversion to actual DraftJS editor state in the middle (as I said before somewhere), that would probably be a bit faster but also a ton of work. This is definitely fast enough for what we use it for.

brianlovin (approved):

brianlovin:
Okay, let's work with it for now! Incremental correctness ftw

PR #4773, opened 2019-03-01
Run e2e tests in parallel

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

If this works, it would also run all the setup and unit tests 9 times, which seems unnecessary to say the least, so we would have to refactor our job setup!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,6 +80,19 @@ jobs:
           root: .
           paths: .
 
+  build_web:
+    <<: *js_defaults
+    steps:
+      - attach_workspace:
+          at: ~/spectrum
+      - run: *setup-and-build-web
+      - run: *build-api
+      - persist_to_workspace:
+        root: .
+        paths:
+          - build-api
+          - build
+
   test_unit:
     <<: *defaults
     docker:
@@ -115,8 +128,6 @@ jobs:
           at: ~/spectrum
       - run: yarn run db:migrate
       - run: yarn run db:seed
-      - run: *setup-and-build-web
-      - run: *build-api
       - run: *start-api
       - run: *start-web
       # Wait for the API and webserver to start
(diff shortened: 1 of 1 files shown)


mxstbr:
Boom this works 😎 

screenshot 2019-03-01 at 11 03 03

2 minute CI runs for e2e tests hell yea

Going to clean this up so we do not run unit tests 9 times and see if I can move the setup to a shared job so that it also only runs once!

mxstbr:
@bahmutov @brian-mann do you know how we can share the database and servers between all the parallel machines instead of spinning them up for each parallel run?

mxstbr:
The latest commit changes it so that we only build hyperion and the API once, and then only run them in each of the parallel containers. That might not work, so if I do not get around to cleaning this up feel free to revert that last commit and ship the rest @brianlovin

mxstbr:
That worked @brianlovin, this is ready!!!! (just flaky e2e tests failing)

An entire CI run now only takes ~7 minutes, down from ~30 minutes!!!!!!! 🎉  

You are going to need to admin-ship this because I renamed the CI jobs, and then you'll have to update the GitHub repo settings to require the new CI jobs, not the old ones to pass.

bahmutov:
First 👏 @mxstbr this looks impressive (and test retries are coming, so flake will be minimized).
Second, there is no general solution to shared DB unfortunately. We would advise to either

- spin the DB in a Docker container with each CircleCI test machine.
- or mock backend calls removing the need for DB (yeah, not ideal)
- splitting tests into "read" and "write" specs and running "read" specs using shared DB, while running "write" specs using separate DBs

I have looked at `test_e2e` job definition and saw that you are running db migrations and seed for every test job, also installing Cypress there. Maybe those steps could move into own job so they all run once and then both the prepared DB image and installed Cypress are ready to go into 9 e2e test jobs?

mxstbr:
@bahmutov I figured as much, thanks for chiming in!

> Maybe those steps could move into own job so they all run once and then both the prepared DB image and installed Cypress are ready to go into 9 e2e test jobs?

That sounds amazing and should speed things up even more, how can I make that happen?!

brianlovin:
Admin-shipping, per @mxstbr's request - will let you two take this discussion further either here or in a separate issue @bahmutov

PR #4820, opened 2019-03-05
Do not run contributors e2e tests 9 times

### LINKED ISSUE #4816: Do not run contributor e2e tests in parallel
See https://circleci.com/gh/withspectrum/spectrum/22782, contributors e2e tests now run on 9 parallel machines—but unfortunately, it runs all of them 9 times 🤦‍♂️ 

We need to disable that and only run them on one machine. We should also reduce the amount of parallelism for our e2e test runs (to 4?), so that multiple builds can run concurrently in parallel and PRs do not block each other.

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Closes #4816

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,8 +119,18 @@ jobs:
       - image: redis:3.2.7
       - image: cypress/base:6
       - image: rethinkdb:2.3.5
-    parallel: true
-    parallelism: 9
+    parallel: |
+      if [ $CYPRESS_RECORD_KEY ]; then 
+        true
+      else
+        false
+      fi
+    parallelism: |
+      if [ $CYPRESS_RECORD_KEY ]; then 
+        4
+      else
+        1
+      fi
     environment:
       TERM: xterm
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,18 +119,8 @@ jobs:
       - image: redis:3.2.7
       - image: cypress/base:6
       - image: rethinkdb:2.3.5
-    parallel: |
-      if [ $CYPRESS_RECORD_KEY ]; then 
-        true
-      else
-        false
-      fi
-    parallelism: |
-      if [ $CYPRESS_RECORD_KEY ]; then 
-        4
-      else
-        1
-      fi
+    parallel: true
+    parallelism: 4
     environment:
       TERM: xterm
     steps:


brianlovin:
@mxstbr how do you want to test this? Just merge into alpha and wait for another contributor PR?

brianlovin (dismissed):
I suppose we'll need this in alpha either way to test things with contributors

mxstbr:
If it still runs parallel for this PR then we are good, I think

brianlovin:
Looks like it's not running at all:

screenshot 2019-03-05 11 12 29

I can't even click on that job, it just spins forever

mxstbr:
Yeah this does not work, one cannot switch the values of a yaml file like that 😢

mxstbr:
@bahmutov how do y'all handle this case?

bahmutov:
Ha, nice catch. So Circle does not pass private env variables to the forked pull requests (which is good behavior). This means one cannot record to the dashboard, which means all 9 machines are going to run the same set of all tests.

Unfortunately, CircleCI does NOT allow you to vary `parallelism:` parameter based on an environment variable, or even conditionally run a job like this https://discuss.circleci.com/t/use-environment-variables-in-parameters/26314

So the only solution I see is to pass parallelism as a parameter to the testing job and use the branch filters to pick this parameter. See example that does this in https://github.com/bahmutov/circleci-parallel-based-on-env

mxstbr:
@bahmutov that does not work for us either as we want all branches submitted by one of us that has the CYPRESS_RECORD_KEY to run in parallel, but not any external branches 😕 

It sucks that Circle does not support this, but oh well. I have simply dialed down the parallelism to 4 for now, but it is really annoying that contributor PRs are going to lock up 4 machines doing the same work. 😕 

This is ready to ship @brianlovin

brianlovin (approved):
Thanks, hopefully Circle will add some better support for this use case in the future

bahmutov:
@mxstbr I have looked at this again - and reworked my example repo https://github.com/bahmutov/circleci-parallel-based-on-env

- all external pull requests are named `pull/<number>` on GitHub

Thus you can use filters to control job parameters based on that: run those branches serially, while running everything else in parallel. See the following examples

- master branch [on two machines](https://circleci.com/workflow-run/39e20855-191d-4da2-8171-675c90e97175)
- internal branch [on two machines](https://circleci.com/workflow-run/b06a7547-b0a1-40d5-aba4-15d154a3657a)
- external pull request from a forked repo [on one machine](https://circleci.com/workflow-run/3f4acfba-e23a-4d77-bb03-a59b4162bc95)

My circleci config at https://github.com/bahmutov/circleci-parallel-based-on-env/blob/master/circle.yml

bahmutov:
haha, I got it: you make _another job_ that requires _both_ parallel and serial jobs. Note that only 1 of those two jobs can run at a time

```yaml
- after-tests:
    requires:
      - "parallel build job"
      - "single build job"
```
You set GitHub branch rule to require "after-tests" job.

screen shot 2019-03-07 at 4 40 04 pm

Again, see the updated circle.yml here https://github.com/bahmutov/circleci-parallel-based-on-env/blob/master/circle.yml

See how it runs here: https://github.com/bahmutov/circleci-parallel-based-on-env/pulls there is internal PR and from a forked commit. Internal runs parallel job, and forked PR runs serial job - but the PR waits for "after-tests" job

## internal pull request branch

"after-tests" job is required but shows 2 checks

screen shot 2019-03-07 at 4 40 45 pm

## forked pull request branch

"after-tests" job is required and shows a single job too

screen shot 2019-03-07 at 4 40 25 pm

mxstbr:
Oh my god @bahmutov you are a wizard!

PR #4825, opened 2019-03-06
Fix /thread/:id modal routes

### LINKED ISSUE #4824: Fix /thread/:id modal routes
Opening a thread with the `/thread/:id` redirect shorthand as a modal redirects to the full thread view. We need to pass the `location.state` onwards in the redirect so that you can also open `/thread/:id` links as a modal.

This currently makes some notification types a bad experience.

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Closes #4824

By passing the location state through `/thread/:id` as a modal works perfectly! :100:

### DIFF
--- a/src/views/thread/container.js
+++ b/src/views/thread/container.js
@@ -218,7 +218,7 @@ class ThreadContainer extends React.Component<Props, State> {
         thread.channel.slug
       }/${slugg(thread.content.title)}~${thread.id}`;
       // $FlowFixMe
-      if (this.props.location.pathname !== properUrl)
+      if (this.props.history.location.pathname !== properUrl)
         // $FlowFixMe
         return this.props.history.replace(properUrl);
     }
--- a/src/views/thread/container.js
+++ b/src/views/thread/container.js
@@ -218,7 +218,7 @@ class ThreadContainer extends React.Component<Props, State> {
         thread.channel.slug
       }/${slugg(thread.content.title)}~${thread.id}`;
       // $FlowFixMe
-      if (this.props.history.location.pathname !== properUrl)
+      if (this.props.location.pathname !== properUrl)
         // $FlowFixMe
         return this.props.history.replace(properUrl);
     }
--- a/src/views/thread/redirect-old-route.js
+++ b/src/views/thread/redirect-old-route.js
@@ -16,10 +16,7 @@ export default getThreadByMatch(props => {
       <Redirect
         to={{
           ...props.location,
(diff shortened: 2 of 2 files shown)


brianlovin:
This breaks with the following repro steps:
- have two recent notifications about someone replying in a thread
- open the first notification - opens the slider correctly
- close
- open the second notification - crashes

screenshot 2019-03-06 10 41 43

mxstbr:
@brianlovin I do not have two notifications locally, does that latest commit fix it?

mxstbr:
Maybe the last commit fixes it, but I cannot repro that locally with normal thread feed cards. Maybe a race condition and/or bad notification data?!

brianlovin:
Hm, I'll test. Just trigger a few notifications for yourself locally with two accounts though ;)

brianlovin:
Still breaks. Error:

screenshot 2019-03-07 08 18 39
screenshot 2019-03-07 08 18 47

Maybe just trigger some fake notifications for yourself and then click around to test locally?

mxstbr:
@brianlovin that bug exists in production today and was not introduced by this PR. Let's ship this because it does make the experience much better, and then work on that bug in a follow-up?

brianlovin:
> that bug exists in production today and was not introduced by this PR

Oh damn, good to know!

brianlovin:
The `thread_spec` test for when a message query parameter exists is failing here. I'll take a look at it now for us

brianlovin:
That e2e test is a legit regression @mxstbr - a url like `http://localhost:3000/spectrum/general/the-first-thread~thread-1?m=MTQ4MzIyNTIwMDAwMQ==` (if you have the local testing db) does break the frontend. Is that because the query thinks that everything including the `?m=...` is part of the thread id?

brianlovin:
It's a minor regression: we still copy/paste the correct urls if you click the message link icon. So maybe this is just an old edge case where a url like `/thread/thread-1?m=MTQ4MzIyNTIwMDAwMQ==` breaks on the redirect...

mxstbr:
Fixed in the latest commit! 👍

mxstbr:
Shipping so this goes out with the next release 👍

brianlovin:
Got it, I'll do a prod cut right now @mxstbr - I know I've said I'd do one the last couple days but this feels like a better time to ship fixes :)

PR #4848, opened 2019-03-09
Fix messages getting empty lines appended

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

We no longer need to append empty lines after embeds now that we have moved editing to plaintext. This also fixes a small logic bug where we would _always_ append empty lines, even if there were no embeds (although that is no longer relevant).


shared/draft-utils/test/__snapshots__/add-embeds-to-draft-js.test.js.snap
```diff
@@ -0,0 +1,43 @@
...
+        },
+      ],
+      "inlineStyleRanges": Array [],
+      "key": "0",
+      "text": " ",
+      "type": "atomic",
+    },
+  ],
+  "entityMap": Object {
+    "-Infinity": Object {
```
brianlovin, line 32: Is this `-Infinity` intentional? Or should that be an actual key number?

mxstbr: Yes that should be 0, I fixed that bug in #4827. Let's merge that first and then update this test?

brianlovin: Merged #4772 - let's make sure the snapshot test fails and then we can update it, otherwise we know the fix is in #4827

brianlovin: Hm, just flow failures now @mxstbr 🤔

mxstbr:
Let's ship this for now as it's very annoying and broken in prod, and then I will fix that test in #4827! @brianlovin 

Going to fix Flow and then we should be ready to go here.

mxstbr:
@brianlovin this should be ready to go!

brianlovin (approved):

PR #4921, opened 2019-03-21
Fix editing messages with automatic embeds

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)

In v3 we hid the links that were embedded, but when editing threads or messages with those hidden links we did not show them again.

Together with a patch to `draft-js-export-markdown` that is live on `convert.spectrum.chat` this now works again for both normal embeds and internal embeds :tada:


brianlovin (approved):
Amazing!

PR #4934, opened 2019-03-22
Serve fake JS file that hard refreshes page if old JS bundle is fetched

mxstbr:
**Status**
- [ ] WIP
- [ ] Ready for review
- [x] Needs testing

- hyperion (frontend)

This might fix our caching issues where folks app breaks after a new deploy. Gotta test on alpha though for sure!
!-- If there are UI changes please share mobile-responsive and desktop screenshots or recordings. -->

### DIFF
--- a/hyperion/index.js
+++ b/hyperion/index.js
@@ -61,6 +61,16 @@ if (process.env.NODE_ENV === 'development') {
   );
 }
 
+// If a file in /static/js is requested, but express.static does not have it,
+// that means the user is on an old version of the app.
+// Serve a piece of JavaScript that hard-reloads the page they are on
+app.get('/static/js/:name', (req, res) => {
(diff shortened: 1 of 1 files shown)


mxstbr:
I could not reproduce the bug on alpha so this is now live in production with deploy https://spectrum-xnwkiyrmvy.now.sh

brianlovin (changes requested):
We should not `window.alert` though - why not just reload?

mxstbr:
I am not sure that there is no edge case where it won't infinitely reload users, so I'd rather they can opt-out of that, if that makes sense 😜 

It's also a bit jarring to use the app and just randomly get a full reload, as we know from our first updating experience (some links are full refreshes if there is an app update).

mxstbr:
Nope.

PR #4958, opened 2019-03-26
Fetch community metadata for user community connection

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

We used to not fetch the community metadata for the user community connection because we did not display any of it unless you visited a community page, which was not very common, and it was quite expensive to calculate.

Now, we actually do display this everywhere and it is cached for a long time so it's not expensive to fetch. By also fetching this from the community connection, the frontend won't do another request to load the community data when switching between communities, heavily reducing the number of API requests we are seeing.

### DIFF
--- a/shared/graphql/queries/user/getUserCommunityConnection.js
+++ b/shared/graphql/queries/user/getUserCommunityConnection.js
@@ -47,7 +47,7 @@ const getUserCommunityConnectionOptions = {
 export const getCurrentUserCommunityConnection = graphql(
   getCurrentUserCommunityConnectionQuery,
   {
-    options: { fetchPolicy: 'cache-and-network' },
+    options: { fetchPolicy: 'cache-first' },
     props: props => ({
       ...props,
       subscribeToUpdatedCommunities: communityIds => {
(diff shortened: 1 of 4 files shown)


mxstbr:
If e2e tests pass I am going to admin-ship this so we can start reducing some of the pressure on our API with the prod cut I am doing @brianlovin!

shared/graphql/queries/user/getUserCommunityConnection.js
```diff
@@ -47,7 +47,7 @@ const getUserCommunityConnectionOptions = {
 export const getCurrentUserCommunityConnection = graphql(
   getCurrentUserCommunityConnectionQuery,
   {
-    options: { fetchPolicy: 'cache-and-network' },
+    options: { fetchPolicy: 'cache-first' },
```
mxstbr, line 50: Reviewers note: now that we have a subscription for this data there is no need to fetch this on every page load. New stuff will come in anyway.

PR #5010, opened 2019-04-01
Add actions dropdown to thread sticky header

### LINKED ISSUE #4949: Add thread actions to sticky thread header

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- hyperion (frontend)

Closes #4949

Makes it possible to mute/unmute and do all other actions from the sticky bar without having to scroll to the top of the messages!

### DIFF
--- a/src/views/thread/components/stickyHeader.js
+++ b/src/views/thread/components/stickyHeader.js
@@ -53,15 +53,7 @@ const StickyHeader = (props: Props) => {
 
       {channel.channelPermissions.isMember && (
         <StickyHeaderActionsContainer>
-          <ActionsDropdown
-            thread={thread}
-            // toggleEdit={this.props.toggleEdit}
-            // lockThread={this.props.threadLock}
-            // isLockingThread={this.props.isLockingThread}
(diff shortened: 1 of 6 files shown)


src/views/thread/components/stickyHeader.js
```diff
@@ -52,6 +53,15 @@ const StickyHeader = (props: Props) => {
 
       {channel.channelPermissions.isMember && (
         <StickyHeaderActionsContainer>
+          <ActionsDropdown
+            thread={thread}
+            // toggleEdit={this.props.toggleEdit}
```
brianlovin: Why are all of these commented out?

mxstbr: All of these moved down to the actionsDropdown and no longer need to be passed through! Going to remove the unnecessary comments

brianlovin (approved):

PR #5050, opened 2019-04-23
Move navigation to community sidebar

mxstbr:
**Status**
- [x] WIP
- [ ] Ready for review
- [ ] Needs testing

- hyperion (frontend)

@brianlovin the only thing missing here is the members list: where should we put it?

Screenshot 2019-04-23 at 10 41 10


brianlovin:
Hey! This is looking great so far - will get a final review in tomorrow and will probably just add some icons in there so that the new items fit in. For members: not sure, will have a think on it - I suppose we could just make it so that clicking the members count or online count opens the members list?

mxstbr:
That's a great idea, I have implemented that in the latest commit! Sounds like the last thing missing here are some icons, wanna add those in?

mxstbr:
Fixing tests!

mxstbr:
Test failure is in the thread_spec, seems like a fluke. Restarting the workflow to see if it fails again!

brianlovin:
Sorry for the slow review here @mxstbr:
- chat item shouldn't be visible if the community doesn't have watercooler enabled
- can we go ahead and add the same notification on/off for the chat "thread" into the list item? This will be the only way for people to mute notifications on chat, which is definitely much needed
- the copy on the tooltip for 'All' is a bit weird - I can't tell if it's because it's so long, wraps onto two lines, or is just worded strangely. The tooltip also only appears if you hover directly over the middle of the list item; doesn't seem to trigger inside the top/bottom padding.

brianlovin:
Working through that list - it's kind of a pain to get the notification controls there for chat, since you'd have to fetch the thread to determine if you are already subscribed or not. That feels like a lot of overhead to add to the page...

brianlovin:
Also the tooltip is hard because that would need to be conditional on both authenticated user and their permissions in the community. E.g. a non-member shouldn't see copy about "channels where you're a member"  - for now just going to remove that tooltip entirely

brianlovin:
Pushed a couple small tweaks, but let me know what you think about how to handle the chat notifications @mxstbr :)

mxstbr:
I think we are fine to load the watercooler thread, because then we can also show an "activity dot" if `watercooler.currentUserLastSeen < watercooler.lastActive`! 😍

mxstbr:
@brianlovin latest commits adds the notification action to the chat list item and clean up the UI across the board a bit.

Unauthenticated and non-member view:

Screenshot 2019-04-26 at 09 46 06

Member view (note the wording change to "All **your** channels"):

Screenshot 2019-04-26 at 09 43 26

I have also added a variable called `newActivity` that we can use the highlight new activity in the chat tab. i was not sure where the put the red dot in the list item, any ideas for that?

We should probably also align the channel view to not have any top tabs, link to the members view when clicking on the count and have the same "search" experience on the posts feed. Going to tackle that now!

mxstbr:
Latest commit aligns the channel view to be consistent with the community view!! 😍😍😍😍

Screenshot 2019-04-26 at 10 45 18

brianlovin:
Holy crap man! Amazing stuff here, all this looks so good. Let me pull the latest and look into where the new activity dot could go on the chat item.

brianlovin:
@mxstbr https://github.com/withspectrum/spectrum/pull/5050/commits/d59587e84fc12ac9cf18e06e27fbb74056e4268b => adds the new activity dot, but intentionally commented it out because we need to perform a separate mutation to make sure that the dot gets cleared out as soon as you view the thread (or at least updates in the background, so that when you click away from chat the dot won't appear there)

Does that feel high pri to you, or should we ship this as-is?

brianlovin:
With the channel view modifications @mxstbr it's no longer possible to view a members list of people in a specific channel. This might be frustrating for people who moderate private channels and want to see who has access to that private channel...

mxstbr:
@brianlovin that should still be visible in the channel settings, right? Maybe for admins instead of showing the "Leave" action in the sidebar we show the settings cog for each channel?

mxstbr:
Working on cleaning this up now! 👍

mxstbr:
Added a settings button to channel list items you are a mod or owner of (at either the community- or channel level):

Screenshot 2019-04-29 at 09 08 11

I don't think it's high prio to fix the click-and-notifications-dot-goes-away, let's do that as a follow-up. Will take care of fixing these e2e tests now and then we should be able to ship?!

brianlovin:
@mxstbr I don't think this is the right change - mods still want to leave channels; we had a lot of complaints back in the day about this and that's why we made it so that admins could leave, but we'd throw the modal in front of the leave dialog just in case. 

I'm not sure the right design here, will think on it today. But in this new design we're sacrificing the ability to manage your channel relationship, get to settings, and view meta info about the channel (members list).

brianlovin (approved):
Let's call this an incremental improvement and go from here.

PR #5052, opened 2019-04-24
Connect to both RethinkDB portals in production

mxstbr:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

- api
- hyperion (frontend)
- athena
- vulcan
- mercury
- hermes
- chronos
- analytics

This should help us not incur downtime in the upcoming planned maintenance.


mxstbr:
I tired deploying this to alpha, but it cannot connect to the second portal. The app is still up though, so this seems to work as we need it to—now I just need to figure out why it's not connecting to the other portal!

mxstbr:
I fixed the secrets and it works! alpha is now running connected to both portals :100:

This is ready to ship and we should get this live for all workers before the planned maintenance, so asap @brianlovin

brianlovin (approved):
Good stuff! Thanks for digging in here :)

PR #5311, opened 2020-01-12
Fix admin site start script

jmgaya:
**Status**
- [ ] WIP
- [X] Ready for review
- [ ] Needs testing

-

Closes the issue raised in [this conversation](https://github.com/withspectrum/spectrum/issues/5310#issuecomment-573406114)


brianlovin (approved):
Thanks for the catch!

PR #5341, opened 2020-04-11
fix code fence newline truncation

### LINKED ISSUE #5165: Feat req: Respect newlines in code snippets
**Issue Type (check one)**

- [ ] Bug Report
- [x] Feature Idea
- [ ] Technical Discussion
- [ ] Question (these will be auto-closed, please ask them on Spectrum instead https://spectrum.chat/spectrum/open)

**Description (type any text below)**

Currently Spectrum destroys empty newlines in code snippets. This is bad because empty newlines are really helpful for readability.

Js-Brecht:
**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

**Deploy after merge**
- hyperion (frontend) _???_

_Not sure if this is the correct answer here._ :confused: 

**Related issues**

Closes #5165

---

This uses an empty theme in place of `undefined` for `prism-react-renderer`, so that it will correctly identify newlines inside of code fences.  The relevant code that makes this identification is [here](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/components/Highlight.js#L87), which will be skipped if the theme is `undefined`.  I haven't noticed any adverse effects by defining an empty theme, but perhaps somebody who is more familiar with that package will have more insight into how far reaching this change could be?

If I were familiar in the least with `prism-react-renderer`, I might even look into setting up an actual theme, instead of using an empty one.  This seems to fix the issue rather effectively, though.

### Before

image

### After

image

### DIFF
--- a/shared/clients/draft-js/renderer/index.js
+++ b/shared/clients/draft-js/renderer/index.js
@@ -72,6 +72,11 @@ const Embed = (props: EmbedData) => {
   return <ExternalEmbed {...props} />;
 };
 
+const EMPTY_THEME = {
+  plain: {},
+  styles: [],
+};
+
 type Options = {
   headings: boolean,
 };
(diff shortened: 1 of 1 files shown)


brianlovin (approved):
Seems like a good fix, thank you!

PR #5401, opened 2020-10-03
Remove data-cy attribute in production

### LINKED ISSUE #5400: Remove "data-cy" attribute in production environment
Why there isn't a babel plugin added (something like **babel-plugin-react-remove-properties)** to remove the **data-cy** attribute in production environment? It would be useful and will reduce bandwidth usage (at least a bit).

angelsbv:
I think title is enough explanation about this PR.

**Status**
- [ ] WIP
- [x] Ready for review
- [ ] Needs testing

Closes #5400 

PD: I am unsure still about what will need to be deployed.

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -163,6 +163,11 @@ module.exports = function override(config, env) {
         },
       })
     );
+    // Remove "data-cy" attribute in production
+    config = injectBabelPlugin(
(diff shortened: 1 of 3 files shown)


brianlovin (approved):
LGTM! Will ship in the future, thank you for looking into this!

